PR #313, opened 2026-07-21
Rewrite container clusters guide: validated GKE DaemonSet + eBPF requirements

maximelb:
## What

Rewrites the **Container Clusters** page (`2-sensors-deployment/endpoint-agent/containers/clusters.md`) based on an end-to-end validated deployment of the sensor as a DaemonSet on GKE Standard, so the page is self-contained for anyone deploying on GKE (or Kubernetes generally).

## Changes

- **glibc image requirement for eBPF**: kernel acquisition is only available for glibc x64 sensors; the page previously showcased an Alpine Dockerfile on a page titled "supporting eBPF" — alpine64 sensors silently run in usermode forever. New sample Dockerfile uses debian:12-slim; Alpine variant documented as usermode-only.
- **debugfs/tracefs mount explained**: it's a separate filesystem invisible through the container's own /sys; without it eBPF tracepoint attach fails and the sensor silently falls back to usermode.
- **containerd netns path** (`/var/run/netns`): the previous docker-only path doesn't exist on modern (containerd) clusters.
- **Installation key via Secret** instead of inline yaml.
- **Scheduling guidance**: custom PriorityClass for utilization-optimized clusters; note that `system-node-critical`/`system-cluster-critical` are quota-restricted outside kube-system on GKE (pod creation fails with a FailedCreate quota error).
- **GKE section**: works as-is on Standard (COS/Ubuntu ship BTF + debugfs); Autopilot unsupported (privileged/hostPath/hostPID/hostNetwork rejected); gVisor caveat.
- **Sensor identity lifecycle**: each pod start enrolls a new SID (expected for elastic fleets); link to Sensor Cull for reaping stale registrations; quota sizing note.
- **How to verify eBPF is active** (`is_kernel_available`, `kernel == true` selector) and the usual causes when it isn't.
- Kept: plain-docker quickstart, non-eBPF variant (now prose instead of a second near-duplicate yaml), SELinux note.

No internal environment details included. markdownlint clean.

https://claude.ai/code/session_013qxDT29gaKi4WpkVuY7Dzb

### DIFF
--- a/docs/2-sensors-deployment/endpoint-agent/containers/clusters.md
+++ b/docs/2-sensors-deployment/endpoint-agent/containers/clusters.md
@@ -1,53 +1,85 @@
 # Container Clusters
 
-You can also run LimaCharlie at the host level in a container cluster system like Kubernetes in order to monitor all running containers on the host with a single Sensor. In fact, this is the preferred method as it reduces the overhead of running LC within every single container.
+You can run LimaCharlie at the host level in a container cluster system like Kubernetes in order to monitor all running containers on the host with a single Sensor. This is the preferred method as it reduces the overhead of running LC within every single container.
 
 This is accomplished by a combination of a few techniques:
 
 1. A privileged container running LC.
-2. LC runs with `HOST_FS` environment variable pointing to the host's root filesystem mounted within the container.
+2. LC runs with the `HOST_FS` environment variable pointing to the host's root filesystem mounted within the container.
 3. LC runs with the `NET_NS` environment variable pointing to the host's directory listing network namespaces.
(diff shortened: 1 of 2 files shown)


lcbill (dismissed):

maximelb:
Follow-up commit: dropped all references to the `refractionpoint/limacharlie_sensor` dockerhub image — it hasn't been updated since May 2023 and its default flavor is CentOS (EOL). Both the clusters page and the Docker installation page now document building your own glibc-based image from a Dockerfile (always gets the latest sensor), and the old alpine "custom image" example on the Docker page was removed since alpine64 has no eBPF support.

lcbill (approved):

PR #319, opened 2026-07-24
docs(cloud-security): onboarding guide for every supported provider

maximelb:
## What

Provider Setup covered **3 of the 13** supported Cloud Security providers (Google Workspace, Cloudflare, AWS). This turns it into a section with a full, step-by-step onboarding guide for **every** platform.

`provider-setup.md` becomes `provider-setup/index.md` — mkdocs renders it at the **same URL**, so no links break and no redirect is needed.

## Per-provider pages

**New:** Google Cloud, Microsoft Azure, Microsoft Entra ID, Okta, 1Password, Auth0, GitHub, OpenAI, Anthropic, LimaCharlie.

**Moved into the section:** AWS, Cloudflare, Google Workspace — content unchanged, each gaining a preflight check-reference table (plus, for Cloudflare, the user-scoped-token surface that covers account members and API-token enumeration).

Every page follows the same structure:

- what the connector collects, and its auth model
- prerequisites (the admin role you need on that platform)
- **required vs optional permissions**, each mapped to the `provider test` check it satisfies and the concrete consequence of skipping it
- click-by-click credential creation, with CLI equivalents where the vendor has one (`gcloud`, `az`, `aws`, `gh`)
- the credential-secret JSON shape and the provider record
- a **full preflight check reference** — every check ID, whether it is required, and what a failure means
- a troubleshooting table keyed on the actual `provider test` error

## Gotchas now documented

These are the ones that cost real onboarding time and were nowhere in the docs:

- **Okta** — the org authorization server *mandates* `private_key_jwt` (client secrets are rejected for `okta.*` scopes); scopes **and** an admin role are both required; ungranted scopes are **silently dropped at mint** rather than erroring; DPoP is locked on for new API Services apps and is handled automatically; `okta.roles.read` is fatal to the *entire* user collector, not just role facets.
- **Azure / Entra** — Graph permissions must be **Application**, not Delegated; sign-in activity additionally needs an Entra ID P1/P2 licence; `azure_subscription_id` is an anchor, not a scope limit.
- **OpenAI** — `api.management.read` must be selected **at key creation**; scope-upgraded keys keep failing.
- **Anthropic** — Console Admin keys carry no scopes at all (stated plainly), and either credential plane can stand alone.
- **1Password** — hosted provisioning vs self-hosted bridge base URLs.
- **LimaCharlie** — `user.ctrl` / `apikey.ctrl` are broader than read-only because no read-only surface exists.

## Also

Every connector in *Connecting Providers* now links to its setup page, and the hub page gains a platform index, the shared-fields reference, and the credential-secret quick reference.

## Verification

- `mkdocs build --strict` passes (exit 0, no warnings) — strict mode catches broken internal links and nav entries
- `markdownlint-cli2` clean across all 360 files
- Permissions, scopes, check IDs, and credential shapes were taken from the collector and preflight implementations rather than from memory; external console steps cross-checked against vendor documentation


lcbill (dismissed):

maximelb:
## Verification pass against official vendor documentation

Re-checked every claim on all 13 pages against the vendors' own docs. Seven corrections, pushed in `dfc7a1f7`.

### Corrections

| Provider | Was | Now |
|---|---|---|
| **Anthropic** | `console.anthropic.com → Organization settings → Admin keys`; Compliance key "request from your account contact" | Console moved to **platform.claude.com → Settings → Admin keys**. Compliance keys are **self-service** at claude.ai → Organization settings → API. Added who can create them, that scopes are **fixed at creation**, that Compliance/Analytics must be enabled first, and the **`read:org_audit`** scope — a single read-only scope built for security-audit integrations, i.e. exactly this connector |
| **Cloudflare** | `My Profile → API Tokens` | That path mints a **user**-owned token. The primary credential is **account**-owned: **Manage Account → API Tokens**. The user path is now correctly attached to the optional `user_api_token` surface |
| **Cloudflare** | Permission "Security Center" | Actual name is **"Security Center Insights"** |
| **Azure / Entra** | `az ad app credential reset --years 2` | That **deletes every existing password and certificate** on the app. Added `--append` and a warning |
| **Auth0** | — | Added a warning against the Management API Explorer's *Create & Authorize a Test Application* button: it grants **every** Management API scope, including the write and secret-bearing ones this connector deliberately avoids |
| **GCP** | "Vertex AI … *(already covered by `roles/viewer`)*" | Could not confirm; claim dropped. Also added why Google's console walkthrough for role recommendations asks for roles beyond what this connector needs |
| **Okta** | DPoP described as vendor behaviour | Okta's service-app guide does not mention DPoP. Reworded as observed behaviour — still accurate, no longer over-claiming |

### Resolved open question

The **self-hosted 1Password SCIM URL** I flagged is settled: 1Password's own troubleshooting guide uses `https://scim.example.com` as the bridge address and explicitly says it "doesn't include directories". So the bridge serves SCIM at the **root**, while hosted provisioning uses `/scim/v2`. Both forms in the table were already right; the page now explains why they differ.

### Confirmed correct, unchanged

- All three **Microsoft Graph application-permission GUIDs**, and the `az ad app permission add --api-permissions {id}=Role` syntax
- All six **Okta scope names**; that `private_key_jwt` is the *only* supported client-auth method for Okta scopes; and that an admin role is required **in addition to** scopes
- Every **GitHub App permission mapping** (members/repos/teams/installations/secrets), against GitHub's endpoint→permission reference
- The **Google Workspace domain-wide-delegation** console path, verbatim
- The **`SecurityAudit` policy ARN** — and its policy document contains every action behind the optional AWS checks (`inspector2:*`, `secretsmanager:ListSecrets`, `rds:Describe*`, `sagemaker:List*`, `bedrock:ListFoundationModels`, `organizations:List*`), corroborating the "no extra policies needed" claim
- **`roles/policyanalyzer.activityAnalysisViewer`** and its three permissions, including the one the collector calls
- The **OpenAI admin-keys URL**

`mkdocs build --strict` and `markdownlint-cli2` remain clean.

lcbill (approved):

PR #328, opened 2026-07-30
docs(cloud-security): the AWS front_doors preflight check

maximelb:
## What was asked

S1 (AWS serverless/edge collectors — legion_cloudsec_host#119) adds a `front_doors` credential
check. Document it, per the rule that a per-provider onboarding scope change lands with the
collector that needs it.

## What was done

Added the `front_doors` row to the AWS `provider test` coverage table in
`docs/cloud-security/provider-setup/aws.md`, plus a note on why it probes the per-function Lambda
reads rather than a representative list call: a missing `lambda:ListFunctionUrlConfigs` grant fails
SILENTLY — every function then reads as having no URL, so a function anyone on the internet can
invoke looks private.

No new onboarding scope is needed, so the setup instructions are unchanged. That was verified two
ways: against both managed policy documents (`SecurityAudit` grants `lambda:List*`,
`lambda:GetPolicy`, `apigateway:GET`, `elasticloadbalancing:Describe*`; `ViewOnlyAccess` grants the
ELBv2 describes and `apigateway:GET`), and live — the `front_doors` check passes against a role
carrying only those two managed policies.

## Testing

Docs-only. Markdown table + admonition, matching the surrounding style.

https://claude.ai/code/session_01L7CkEf1fEa9Y8dBn5Utrx1

### DIFF
--- a/docs/cloud-security/provider-setup/aws.md
+++ b/docs/cloud-security/provider-setup/aws.md
@@ -129,13 +129,18 @@ With `SecurityAudit` + `ViewOnlyAccess`, every optional surface above also
 passes — no extra policies needed.
 
 !!! note "Why `front_doors` probes more than one call"
-    A missing `lambda:ListFunctionUrlConfigs` grant does not fail loudly: every
-    function simply reads as having no URL, so a function that anyone on the
-    internet can invoke looks private. The check therefore exercises the real
-    per-function reads against one existing function rather than a representative
-    list call. Both `SecurityAudit` (`lambda:List*`, `lambda:GetPolicy`,
-    `apigateway:GET`, `elasticloadbalancing:Describe*`) and `ViewOnlyAccess` grant
-    everything it needs.
+    Lambda's exposure verdict needs three separate grants —
+    `lambda:ListFunctions`, `lambda:ListFunctionUrlConfigs` and
+    `lambda:GetPolicy` — and a role missing any one of them still passes a
+    single representative list call. The check therefore exercises the real
+    per-function reads against one existing function, so a hand-rolled
+    least-privilege policy is caught at connect time instead of leaving the
+    Lambda inventory unreadable on every sweep.
+
+    The two managed policies grant these BETWEEN them, not each on its own:
+    `SecurityAudit` supplies `lambda:List*` and `lambda:GetPolicy`,
+    `ViewOnlyAccess` the ELBv2 describes, and both supply `apigateway:GET`.
+    Attach both, as the setup above does.
 
 ## Organization guardrails (SCPs and RCPs)
 
--- a/docs/cloud-security/provider-setup/aws.md
+++ b/docs/cloud-security/provider-setup/aws.md
@@ -122,10 +122,20 @@ the AWS-specific checks follow.
 | `inspector` | — | Workload vulnerability findings unavailable. |
 | `secrets_manager` | — | Secret-store inventory unavailable. |
 | `data_stores` | — | RDS / DynamoDB / Redshift inventory unavailable. |
+| `front_doors` | — | Lambda / API Gateway / load-balancer inventory unavailable — a public function URL, API stage or internet-facing load balancer stays invisible to the exposure model. |
 | `ai_services` | — | SageMaker / Bedrock inventory unavailable. |
 
 With `SecurityAudit` + `ViewOnlyAccess`, every optional surface above also
 passes — no extra policies needed.
+
+!!! note "Why `front_doors` probes more than one call"
+    A missing `lambda:ListFunctionUrlConfigs` grant does not fail loudly: every
+    function simply reads as having no URL, so a function that anyone on the
+    internet can invoke looks private. The check therefore exercises the real
+    per-function reads against one existing function rather than a representative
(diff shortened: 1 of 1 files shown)


lcbill (dismissed):

docs/cloud-security/provider-setup/aws.md
```diff
@@ -122,11 +122,21 @@ the AWS-specific checks follow.
...
 passes — no extra policies needed.
 
+!!! note "Why `front_doors` probes more than one call"
+    A missing `lambda:ListFunctionUrlConfigs` grant does not fail loudly: every
+    function simply reads as having no URL, so a function that anyone on the
+    internet can invoke looks private. The check therefore exercises the real
+    per-function reads against one existing function rather than a representative
+    list call. Both `SecurityAudit` (`lambda:List*`, `lambda:GetPolicy`,
+    `apigateway:GET`, `elasticloadbalancing:Describe*`) and `ViewOnlyAccess` grant
+    everything it needs.
```
maximelb: `ViewOnlyAccess` does **not** grant everything this check needs. Verified against both current default policy versions today:

- `SecurityAudit` — `lambda:List*`, `lambda:GetPolicy`, `apigateway:GET`, `elasticloadbalancing:Describe*` ✅ (as written)
- `arn:aws:iam::aws:policy/job-function/ViewOnlyAccess` — `lambda:List*`, `apigateway:GET`, `elasticloadbalancing:Describe{LoadBalancers,Listeners,TargetGroups,TargetHealth,LoadBalancerAttributes,InstanceHealth}`. **No `lambda:GetPolicy`.**

The probe calls `lambda:GetPolicy` and fails the check when it is denied, so "Both … grant everything it needs" is wrong on the natural reading (each individually). It only holds for the *union*, which is what the setup page attaches. A customer building a hand-rolled least-privilege role from this sentence will produce one that fails `front_doors`. Suggest: "`SecurityAudit` grants all of them; `ViewOnlyAccess` adds nothing this check needs beyond what `SecurityAudit` already covers."

docs/cloud-security/provider-setup/aws.md
```diff
@@ -122,11 +122,21 @@ the AWS-specific checks follow.
...
 | `data_stores` | — | RDS / DynamoDB / Redshift inventory unavailable. |
+| `front_doors` | — | Lambda / API Gateway / load-balancer inventory unavailable — a public function URL, API stage or internet-facing load balancer stays invisible to the exposure model. |
 | `ai_services` | — | SageMaker / Bedrock inventory unavailable. |
 
 With `SecurityAudit` + `ViewOnlyAccess`, every optional surface above also
 passes — no extra policies needed.
 
+!!! note "Why `front_doors` probes more than one call"
+    A missing `lambda:ListFunctionUrlConfigs` grant does not fail loudly: every
+    function simply reads as having no URL, so a function that anyone on the
```
maximelb: This describes a failure mode the collector deliberately does **not** have, so it tells the customer the wrong thing about what a missing grant costs them.

A denied `ListFunctionUrlConfigs` does not read as "no URL". `mapAWSError` leaves `AccessDenied` unmapped ⇒ fatal ⇒ `collect_frontdoor.go:117-123` returns a `*collect.CoverageError`, poisoning `ComputeInstance` + `HasPermission` for the account. The paired PR's own test asserts exactly that: `TestLambdaCollectorErrorTaxonomy` — "a denied FunctionURLs read must poison coverage rather than report no door". So functions do not look private; the Lambda inventory simply stops being authoritative (rows frozen, no deletes) and the operator sees nothing in the product either way.

The probe is still worth having, and the *structural* argument still stands — `lambda:ListFunctions` can be granted while `ListFunctionUrlConfigs` is not, so a representative-list-only check would pass and every sweep would then fail — it's just a different consequence. Recommend restating it as: a list-only probe would pass on a role that cannot read function URLs, and the sweep would then leave Lambda coverage permanently incomplete. (The same wording appears in `producer/aws/test.go`'s probe comment and in both PR bodies.)

maximelb:
Open findings (inline threads above):

- `ViewOnlyAccess` does not grant `lambda:GetPolicy` (verified against the current default policy version), so "Both … grant everything it needs" is wrong per-policy — only the union holds.
- The stated failure mode is not the code's: a denied `ListFunctionUrlConfigs` poisons the account's Lambda coverage (`CoverageError`), it does not make functions read as having no URL. Same wording is in `producer/aws/test.go` and both PR bodies.

lcbill (approved):

docs/cloud-security/provider-setup/aws.md
maximelb: Fixed in e556f763. You are right — ViewOnlyAccess carries lambda:List* but not lambda:GetPolicy. The note now says the two policies grant these BETWEEN them and attributes each grant to the policy that actually supplies it.

docs/cloud-security/provider-setup/aws.md
maximelb: Fixed in e556f763 (and in the same commit's producer/aws/test.go comment on the host PR). The collector poisons its Lambda coverage on a denied read rather than reporting functions as private, so the honest framing of the check's value is catching a hand-rolled least-privilege policy at connect time instead of leaving the Lambda inventory unreadable on every sweep — which is what both now say.

PR #329, opened 2026-07-30
docs(cloud-security): GCP serverless scopes — Cloud Run + Cloud Functions

maximelb:
## What was asked

Cloud Run and Cloud Functions inventory is now collected for Google Cloud connections, together
with a verdict on whether each workload can be invoked from the internet with no authentication.
Document the scopes that verdict needs — before a customer discovers them as a silent denial.

## What was done

All in `docs/cloud-security/provider-setup/gcp.md`.

- **The required baseline already covers it.** `roles/viewer` + `roles/iam.securityReviewer`
  grant both the list and the `getIamPolicy` reads, so nothing changes for a customer following
  the standard setup. Said explicitly, so the two new rows are not misread as new requirements.
- **Two optional rows** — `roles/run.viewer` and `roles/cloudfunctions.viewer` — for the
  least-privilege alternative (`roles/browser` plus per-service viewers), where they are what
  turns serverless coverage on. Added to the setup script too.
- **Both halves of each grant are called out.** Without `getIamPolicy` we can list a service but
  cannot tell whether anyone on the internet can invoke it, and we report **no verdict** rather
  than guess — so serverless exposure findings are *absent, not empty*, which is a different
  thing from "you have none".
- **The 2nd-gen trap.** A 2nd-gen Cloud Function runs on Cloud Run and its *invoker* permission
  lives on the underlying Cloud Run service (`roles/run.invoker`), not on the function
  (`roles/cloudfunctions.invoker` is the 1st-gen role). So a grant with
  `roles/cloudfunctions.viewer` but **not** `roles/run.viewer` lists 2nd-gen functions while
  leaving every one of them without a public-access verdict.
- **New `serverless` row** in the preflight-check table, and two troubleshooting rows — including
  the symptom that looks like good news: services listed, none ever flagged public, because
  `getIamPolicy` is denied.
- `run.googleapis.com` + `cloudfunctions.googleapis.com` added to the enable list; the page intro
  now mentions serverless alongside compute.

## Testing

- Verified against the public Google Cloud IAM and Cloud Functions documentation: the permission
  names (`run.services.list`, `run.services.getIamPolicy`, `cloudfunctions.functions.list`,
  `cloudfunctions.functions.getIamPolicy`), the role names, and the 2nd-gen invoker behaviour
  (the Cloud Functions access-control page documents granting `roles/run.invoker` to `allUsers`
  via `gcloud run services add-iam-policy-binding`).
- Markdown-only change: table rows plus three admonitions and two script lines. No numbered lists
  touched, so the list-numbering check is unaffected. `mkdocs` is not installed locally, so the
  rendered-site checks run in CI.

https://claude.ai/code/session_01L7CkEf1fEa9Y8dBn5Utrx1

### DIFF
--- a/docs/cloud-security/provider-setup/gcp.md
+++ b/docs/cloud-security/provider-setup/gcp.md
@@ -115,11 +115,15 @@ Each adds one inventory or analysis surface. Skipping one leaves that surface
 !!! note "2nd-gen functions are authorized as Cloud Run services"
     A 2nd-gen Cloud Function runs on Cloud Run, and its **invoker** permission
     lives on the underlying Cloud Run service (`roles/run.invoker`) rather than
-    on the function (`roles/cloudfunctions.invoker` is the 1st-gen role). A
-    least-privilege grant with `roles/cloudfunctions.viewer` but **not**
-    `roles/run.viewer` will therefore list your 2nd-gen functions while leaving
-    every one of them without a public-access verdict. If you use 2nd-gen
-    functions, grant both.
+    on the function (`roles/cloudfunctions.invoker` is the 1st-gen role). So
+    assessing a 2nd-gen function's public access reads the **Cloud Run** side,
+    not the Cloud Functions side.
+
+    If you are assembling a least-privilege grant rather than using the required
+    baseline, grant both `roles/run.viewer` and `roles/cloudfunctions.viewer` and
+    let `provider test` confirm it: the `serverless` check exercises both APIs and
+    reports each separately, so it will tell you if one half is missing rather
+    than leaving you to reason about role contents.
 
 !!! note "`osconfig_vuln` does not prove the inventory join"
     The `osconfig_vuln` check probes the vulnerability-report read only, so it
@@ -270,6 +274,6 @@ the GCP-specific checks follow.
 | `iam` fails on `getIamPolicy` | `roles/viewer` alone does not cover every `getIamPolicy` | Add `roles/iam.securityReviewer` |
 | A check passes with *"API not enabled on the probed project"* | Benign — the sweep skips API-disabled projects | Enable the named API if you want that surface; otherwise ignore |
 | `activity_ciem` fails with *Recommender API not enabled* | Recommender / Policy Analyzer not enabled on the probed project | Enable `recommender.googleapis.com` and `policyanalyzer.googleapis.com` |
-| `serverless` fails on `run` only | The grant covers Cloud Functions but not Cloud Run | Add `roles/run.viewer`. 2nd-gen functions are authorized as Cloud Run services, so without it they list with no public-access verdict |
+| `serverless` fails on `run` only | The grant reaches Cloud Functions but not Cloud Run | Add `roles/run.viewer`. 2nd-gen functions are authorized as Cloud Run services, so without that read they list with no public-access verdict |
 | Cloud Run services appear but none is ever flagged public | `getIamPolicy` is denied on Cloud Run, so the invoker verdict is unobserved rather than negative | Add `roles/iam.securityReviewer` (or `roles/run.viewer`) and re-run the sweep |
 | Inventory is missing whole projects | Those projects are not `ACTIVE`, or the grant is on a narrower node | Confirm project state, and that the binding is at the scope you configured |
--- a/docs/cloud-security/provider-setup/gcp.md
+++ b/docs/cloud-security/provider-setup/gcp.md
@@ -104,6 +104,14 @@ Each adds one inventory or analysis surface. Skipping one leaves that surface
     verdict rather than guess — so serverless exposure findings are **absent,
     not empty**, which is a different thing from "you have none".
 
+!!! note "Services published only through a load balancer"
+    A Cloud Run service or function whose ingress is **internal and Cloud Load
+    Balancing** cannot be reached at its own `run.app` URL, but it *can* be
+    published to the internet through an external load balancer in front of it.
+    We do not yet collect load balancers, so such a service is inventoried but
+    is **not** reported as internet-facing. Services reachable directly at their
+    own URL are assessed normally.
+
 !!! note "2nd-gen functions are authorized as Cloud Run services"
     A 2nd-gen Cloud Function runs on Cloud Run, and its **invoker** permission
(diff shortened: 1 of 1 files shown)


lcbill (dismissed):

docs/cloud-security/provider-setup/gcp.md
```diff
@@ -85,8 +87,32 @@ Each adds one inventory or analysis surface. Skipping one leaves that surface
...
 
+!!! note "Serverless already works on the required baseline"
+    The required `roles/viewer` + `roles/iam.securityReviewer` pair **already**
+    grants both the list and the `getIamPolicy` reads for Cloud Run and Cloud
+    Functions, so you do not need to add anything for serverless coverage. The
+    two roles above exist for the least-privilege alternative (`roles/browser`
+    plus per-service viewers), where they are what turns serverless coverage on.
+
+    Both halves of each grant matter, and the second is the one that gets
+    missed: without `getIamPolicy` we can list a service but cannot tell whether
```
maximelb, line 102: **"we can list a service but cannot tell whether it is invocable" is not what the collector does today — a denied `getIamPolicy` suppresses the INVENTORY too.**

In `legion_cloudsec_host` the serverless collectors declare `ComputeInstance` as their only type, and an unreadable invoker policy returns `CoverageError{Uncovered: ["ComputeInstance"]}`. `producer/bulk.go` skips an uncovered sub-scope for upserts as well as deletes, so nothing is written: with `list` granted and `getIamPolicy` denied the customer sees **no Cloud Run / Cloud Functions rows at all**, not rows without a verdict. So this paragraph, and the troubleshooting row at L266 ("Cloud Run services appear but none is ever flagged public"), describe a state that cannot occur.

I have raised the underlying behaviour on the host PR — whichever way that lands, these two passages need to match it, because they are the text a customer uses to decide whether a partial grant is acceptable. If the collector is changed to keep the inventory covered, the wording here becomes correct as written; if not, it should say the tier is not collected at all without `getIamPolicy`.

maximelb:
Open finding (details inline):

- `gcp.md` L102 + L266 — both describe "listed but no public-access verdict" when `getIamPolicy` is denied; the collector actually returns partial coverage on its only type, which suppresses the serverless inventory entirely (`bulk.go` skips uncovered sub-scopes for upserts too). Raised on the host PR; the copy needs to match whichever behaviour lands.

No public-repo hygiene issues found (no customer names, internal repo names, internal URLs, or org ids); the role/permission table and the gen2-`run.invoker` note match the collector.

lcbill (dismissed):

docs/cloud-security/provider-setup/gcp.md
maximelb: Fixed in 7bb8fe65 — and the fix was in the collector, not the page. You were right that the state was unreachable, so the collector was changed to make it the real behaviour: an unreadable invoker policy now degrades the ROW (inventoried, no verdict, InvokerPolicyUnread set) instead of poisoning coverage, precisely because poisoning cost the whole project's serverless inventory. The page's "listed but no verdict" wording is now accurate. Separately added an admonition for the load-balancer-only case, which is a state customers will actually hit and was undocumented.

docs/cloud-security/provider-setup/gcp.md
```diff
@@ -85,8 +87,40 @@ Each adds one inventory or analysis surface. Skipping one leaves that surface
...
+    own URL are assessed normally.
+
+!!! note "2nd-gen functions are authorized as Cloud Run services"
+    A 2nd-gen Cloud Function runs on Cloud Run, and its **invoker** permission
+    lives on the underlying Cloud Run service (`roles/run.invoker`) rather than
+    on the function (`roles/cloudfunctions.invoker` is the 1st-gen role). A
+    least-privilege grant with `roles/cloudfunctions.viewer` but **not**
+    `roles/run.viewer` will therefore list your 2nd-gen functions while leaving
+    every one of them without a public-access verdict. If you use 2nd-gen
+    functions, grant both.
```
maximelb: **This note tells customers something that is not true of the GCP roles.** `roles/cloudfunctions.viewer` already includes `run.services.list`, `run.services.getIamPolicy` and `run.locations.list` — verified against the live IAM API (`GET https://iam.googleapis.com/v1/roles/cloudfunctions.viewer`). Cloud Functions Viewer was extended with the Cloud Run reads precisely *because* gen2 runs on Cloud Run.

So a least-privilege grant with `roles/cloudfunctions.viewer` and **not** `roles/run.viewer` does get a public-access verdict on 2nd-gen functions; the stated trap does not exist, and "if you use 2nd-gen functions, grant both" over-grants on a false premise. What `run.viewer` actually buys on top is standalone Cloud Run *service* inventory (functions.viewer alone gives the reads but the collector still needs `cloudfunctions.functions.list` to be the one enumerating).

The troubleshooting row below ("`serverless` fails on `run` only → the grant covers Cloud Functions but not Cloud Run") inherits the same error. Same premise is repeated in `legion_cloudsec_host/docs/GCP-SERVERLESS.md` ("the trap that a functions-only grant lists gen2 functions with no public-access verdict") and in the `resolveBackingServices` comment ("a cloudfunctions-viewer-without-run-viewer grant — a configuration the onboarding docs explicitly support"). The once-per-region indexing is still right on its own merits; only the justification and this customer-facing advice need correcting.

lcbill (approved):

docs/cloud-security/provider-setup/gcp.md
maximelb: Fixed in 58595e6f, though not quite as proposed — I could not reproduce your role-contents reading. Querying `roles/cloudfunctions.viewer` and `roles/run.viewer` with `view=FULL` returned zero includedPermissions for both from this credential, so I cannot confirm either that cloudfunctions.viewer carries the Cloud Run reads or that run.viewer carries run.services.getIamPolicy. Rather than assert a role-contents claim in either direction, the note now states only the mechanism that is definitely true (a 2nd-gen function's invoker permission lives on the Cloud Run service, so assessing it reads the Cloud Run side) and points at `provider test` as the authority: its serverless check exercises both APIs and reports each separately, so an operator learns which half is missing instead of reasoning about role contents. The least-privilege advice to grant both is unchanged, since it is correct either way.

PR #336, opened 2026-08-03
docs(cloud-security): document custom posture rules and remediation SLAs

maximelb:
## What was asked

Two customer-facing Cloud Security features are live and undocumented in the public docs: the `rules` and `sla` policy types on the `cloudsec_policy` hive. Document both, matching the existing section's structure and tone, with every JSON example validated against the real hive write gate before it ships.

## What was done

**Two new pages**

- `docs/cloud-security/custom-rules.md` — **Custom Posture Rules** (`policy_type: "rules"`). What a rule is (a real D&R detection over one canonical resource, presented as a `cloud_resource.<type>` event — not a new DSL); the record schema; why rule ids are permanent (fingerprint input *and* compliance-catalog join key) and why they are namespaced `custom-`; the operator allowlist as a safety boundary; the rejected value syntaxes (`[[name]]`, `{{ … }}`) and wildcard paths; the `scope` cross-element trap with a worked firewall example; "never fire on silence"; override precedence; how evaluation works (stateless per resource — a new rule applies to the whole existing estate, findings close by themselves, they land in the normal worklist with full lifecycle); the compliance interaction; the bounds and the reject-per-record / truncate-across-records stance; and an explicit "what this is not" (no cross-resource or temporal correlation, no console editor).
- `docs/cloud-security/remediation-sla.md` — **Remediation SLAs** (`policy_type: "sla"`). Why there is no default SLA and why that is deliberate; the starter policy; the record schema and match semantics (ordered, first-match-wins, AND across keys / OR within a key, empty `match` as a deliberate catch-all); `due_days: 0` as an exemption rather than "due immediately"; stored (`due_at`, `sla_source`) vs derived (`sla_state`) and why the state cannot be materialized; `due_soon` as the final quarter of the finding's own window; due dates anchoring to first-seen; exempt-not-paused; breach events; the bounds; and the worklist / CLI / API surface.

**Existing pages extended**

| Page | Change |
|---|---|
| `configuration.md` | `cloudsec_policy` purposes row updated; new `rules` and `sla` policy-type sections |
| `findings.md` | SLA fields on the finding, a **Due dates** section, `cloud_finding.sla_breached`, and a note that self-authored rules produce ordinary findings |
| `cli.md` | `--sla` and `--sort due_at`, plus the note that both are Hive-authored records rather than missing commands |
| `api-reference.md` | `sla[]` / `owner[]` selectors and the `due_at` sort key, the finding's SLA fields, and the breach event |
| `automation.md` | IaC recipes for both records and an **Escalating an SLA breach** D&R rule |
| `compliance.md` | the disabled-rule case added to the `NOT_ASSESSED` definition |
| `index.md`, `mkdocs.yml` | capability rows, documentation list, nav |

**Two things worth a reviewer's attention**

- The CLI's `--sla` / `--sort due_at` are merged but **not in a released version** — 5.6.1 is the newest on PyPI and does not carry them. The CLI page says so explicitly and points at the REST parameters in the meantime; that note should be deleted once a release ships.
- The API-reference findings row previously omitted `owner[]`, which the gateway has been forwarding; it is listed alongside `sla[]` now.

## Testing

**Every JSON example was validated against the live hive write gate** on a test organization, in the exact form it appears in the docs — all 8 records (4 rule examples, 2 SLA examples, 2 Automation recipes) extracted programmatically from the rendered pages and written back. All accepted; the validation record was deleted afterwards and the organization's pre-existing records were left untouched and verified intact.

The six rejection messages quoted in the docs are the gate's real output, captured verbatim (missing `custom-` prefix, wildcard path, disallowed operator, `subject_path` written as a detection path, non-authorable `finding_class`, no-op override). The operator allowlist published in the docs is the one the gate itself prints in that error.

Repo checks:

- `mkdocs build --strict` — clean
- `python scripts/check-list-numbering.py docs site --baseline …` — 0 new numbering breaks
- `pytest tests/` — 57 passed

https://claude.ai/code/session_01FyyEPjqXMJWRWtTnBjchnS

### DIFF
--- a/docs/cloud-security/api-reference.md
+++ b/docs/cloud-security/api-reference.md
@@ -40,8 +40,8 @@ Shared behaviors:
 
 | Route | Returns |
 |---|---|
-| `GET /findings` | `{findings, next_cursor}` — the risk-ranked worklist. Filters: `severity[]`, `finding_class[]`, `status[]`, `account[]`, `reachable`, `kev`, `q`, `sort`, `order`, `cursor`, `limit`. |
-| `GET /findings/facets` | `{facets}` — cross-filtered facet counts under the same selectors. |
+| `GET /findings` | `{findings, next_cursor}` — the risk-ranked worklist. Filters: `severity[]`, `finding_class[]`, `status[]`, `account[]`, `owner[]`, `sla[]`, `reachable`, `kev`, `q`, `sort`, `order`, `cursor`, `limit`. `sort` is `lc_risk` (default), `severity`, `first_seen`, or `due_at`; `due_at` is the one key that defaults to **ascending** and sorts findings with no due date last. |
+| `GET /findings/facets` | `{facets}` — cross-filtered facet counts under the same selectors, including an `sla` dimension counting each state. |
 | `GET /findings/classes` | `{classes}` — the canonical `finding_class` enum (the 11 values), for building selectors. |
 | `GET /findings/causes` | `{causes, distinct}` — findings grouped by their **cause**, the mutable object (a firewall rule, a role binding) whose single edit resolves all of them. Takes the full findings selectors, plus `cause` for the exact count of one cause and `limit` (default 20, max 200) for the top causes; `distinct` is the total number of causes matching the filter. |
 | `GET /findings/{finding_id}` | `{finding}` — one finding in full. |
@@ -160,6 +160,7 @@ carry the same shape) — key fields:
 | `runtime_sids` | Sensors resolved onto the affected asset. |
 | `status`, `resolution`, `resolved_by`, `resolution_expires_at`, `owner`, `ticket` | The triage overlay. |
 | `first_seen`, `last_seen` | Lifecycle timestamps. |
+| `due_at`, `sla_source`, `sla_state` | The remediation clock. `due_at` (the deadline) and `sla_source` (the policy clause that set it, or `default`) are stored; `sla_state` — `breached` \| `due_soon` \| `on_track` \| `exempt` \| `none` — is **derived on every read** from `due_at`, `first_seen`, and `status`, because it changes with the clock and nothing writes when a deadline passes. All three are absent until the organization writes an [`sla` policy](remediation-sla.md); there is no default SLA. |
 
 ## Events
 
@@ -182,6 +183,11 @@ summary still counts the whole estate).
 - `cloud_finding.still_open` — re-asserted at most once a day for open
   findings with a linked ticket (the case-sync verb); the payload carries the
   `ticket` alongside the identity fields.
+- `cloud_finding.sla_breached` — emitted once, on the first projection pass
+  that observes an open finding past its `due_at`. The breach is latched, so
+  the event never re-fires for the same breach of the same deadline. Requires
+  an [`sla` policy](remediation-sla.md); granularity is the reprojection
+  cadence, not the second.
 
 **Disposition** — flat payload
 `{finding_id, fingerprint, finding_class, actor, note?}`, plus the finding's
--- a/docs/cloud-security/api-reference.md
+++ b/docs/cloud-security/api-reference.md
@@ -160,7 +160,7 @@ carry the same shape) — key fields:
 | `runtime_sids` | Sensors resolved onto the affected asset. |
 | `status`, `resolution`, `resolved_by`, `resolution_expires_at`, `owner`, `ticket` | The triage overlay. |
 | `first_seen`, `last_seen` | Lifecycle timestamps. |
-| `due_at`, `sla_source`, `sla_state` | The remediation clock. `due_at` (the deadline) and `sla_source` (the policy clause that set it, or `default`) are stored; `sla_state` — `breached` \| `due_soon` \| `on_track` \| `exempt` \| `none` — is **derived on every read** from `due_at`, `first_seen`, and `status`, because it changes with the clock and nothing writes when a deadline passes. All three are absent until the organization writes an [`sla` policy](remediation-sla.md); there is no default SLA. |
+| `due_at`, `sla_source`, `sla_state` | The remediation clock. `due_at` (the deadline) and `sla_source` (the policy clause that set it, or `default`) are stored, and are **absent** until the organization writes an [`sla` policy](remediation-sla.md) that covers the finding — there is no default SLA. `sla_state` is always present: it is **derived on every read** from `due_at`, `first_seen`, and `status` (because it changes with the clock and nothing writes when a deadline passes) and reads `none` for any finding no clause covers. Values: `breached` \| `due_soon` \| `on_track` \| `exempt` \| `none`. |
 
 ## Events
 
--- a/docs/cloud-security/compliance.md
+++ b/docs/cloud-security/compliance.md
@@ -39,11 +39,11 @@ The report is per-control, and each control lands in one of four states:
   the manual / attestation controls (policy documents, board sign-off,
   interview evidence): the framework marks them as not auto-assessable, so no
   rule is run and no verdict is invented. A control also lands here when the
-  detection that would have decided it never ran — including when **you have
-  disabled every rule that evidences it** through a
-  [`rules` policy](custom-rules.md#compliance-interaction). A zero-violation
-  result from a detector that did not run is not evidence of compliance, so it
-  is never reported as a PASS.
+  detection that would have decided it never ran — including when a
+  [`rules` policy](custom-rules.md#compliance-interaction) has **disabled every
+  rule the control relies on** and it has no other basis to be graded from. A
+  zero-violation result from a detector that did not run is not evidence of
+  compliance, so it is never reported as a PASS.
 - **NOT_APPLICABLE** — the control has nothing to assess in *this* estate:
   either the framework's cloud has no resources connected, or the control
   declares no resource types to apply to.
--- a/docs/cloud-security/compliance.md
+++ b/docs/cloud-security/compliance.md
@@ -38,7 +38,12 @@ The report is per-control, and each control lands in one of four states:
 - **NOT_ASSESSED** — the control cannot be evaluated automatically. These are
   the manual / attestation controls (policy documents, board sign-off,
   interview evidence): the framework marks them as not auto-assessable, so no
-  rule is run and no verdict is invented.
+  rule is run and no verdict is invented. A control also lands here when the
+  detection that would have decided it never ran — including when **you have
+  disabled every rule that evidences it** through a
+  [`rules` policy](custom-rules.md#compliance-interaction). A zero-violation
+  result from a detector that did not run is not evidence of compliance, so it
+  is never reported as a PASS.
 - **NOT_APPLICABLE** — the control has nothing to assess in *this* estate:
   either the framework's cloud has no resources connected, or the control
   declares no resource types to apply to.
--- a/docs/cloud-security/custom-rules.md
+++ b/docs/cloud-security/custom-rules.md
@@ -1,0 +1,493 @@
+# Custom Posture Rules
+
+The built-in CSPM pack is not the ceiling. A `rules`-typed
+[`cloudsec_policy`](configuration.md#cloudsec_policy) record lets an organization
+author its **own** posture detections — and retune, re-severity, or switch off the
+built-in ones — in exactly the same format the built-in pack uses.
+
+There is no separate rule language to learn: a posture rule is a **real D&R
+detection**, in the same `op` / `path` / `value` dictionary vocabulary as the
+detection & response rules you already write, evaluated against a cloud resource
+presented as a `cloud_resource.<ResourceType>` event whose body is the resource's
+normalized properties.
+
+!!! info "Authored as Hive records"
+    Unlike classification, coverage, exclusions, and suppression, custom rules
+    have **no console editor**. They are written as Hive JSON through
+    `limacharlie hive set` (or the Hive API), which also makes them the easiest
+    policy to keep in a git repository and push to a fleet — see
+    [Automation & IaC](automation.md#custom-posture-rules).
+
+## A first rule
+
+```json
+{
+  "policy_type": "rules",
+  "rules": {
+    "rules": [
+      {
+        "id": "custom-public-bucket-outside-cdn",
+        "name": "Public bucket outside the CDN convention",
+        "resource_type": "DataStore",
+        "finding_class": "public_exposure",
+        "severity": "HIGH",
+        "title": "Storage bucket is readable by anyone on the internet",
+        "detect": {
+          "op": "and",
+          "rules": [
+            {"op": "is", "path": "event/store_kind", "value": "bucket"},
+            {"op": "is", "path": "event/is_public", "value": true},
+            {"op": "starts with", "path": "event/name", "value": "cdn-", "not": true}
+          ]
+        },
+        "meta": {
+          "description": "A bucket granting read access to anyone, outside the cdn- naming convention used for deliberately public assets.",
+          "rationale": "Anonymous read on a bucket exposes every object in it, and our public assets are all published under the cdn- prefix.",
+          "references": ["https://cloud.google.com/storage/docs/access-control/making-data-public"],
+          "false_positives": "A newly created public asset bucket that has not been renamed to the cdn- convention yet."
+        }
+      }
+    ]
+  }
+}
+```
+
+```bash
+limacharlie hive set --hive-name cloudsec_policy --key my-rules \
+  --oid $OID --input-file rules.json --enabled
+```
+
+The record is validated **synchronously**. The Hive compiles every `detect` block
+on the real detection engine and checks every vocabulary, so a rule that could
+never run is rejected while you are still looking at the error — never accepted
+and then silently skipped hours later inside a projection you cannot see.
+
+An organization may hold **many** `rules` records; they compose into one policy in
+record-name order.
+
+## The record
+
+`rules` is an object with two independent lists, `rules` (your own detections) and
+`overrides` (retunes of rules that already exist). At least one must be non-empty
+— a record that adds nothing and overrides nothing is rejected rather than saved
+as a no-op.
+
+### Rule fields
+
+| Field | Required | Meaning |
+|---|---|---|
+| `id` | ✔ | Stable identifier, kebab-case, **must start with `custom-`**, max 64 characters. Never rename it — see [Rule ids are permanent](#rule-ids-are-permanent). |
+| `resource_type` | ✔ | Which resources the rule is evaluated against; routes it to `cloud_resource.<type>` events. |
+| `finding_class` | ✔ | `misconfig` or `public_exposure` — the whole authorable set (see below). |
+| `severity` | ✔ | `CRITICAL` \| `HIGH` \| `MEDIUM` \| `LOW` \| `INFO`. |
+| `title` | ✔ | What the operator reads on the finding. State the **risk**, not the setting: "Bucket is readable by anyone on the internet", never "uniform bucket-level access is disabled". Max 256 characters. |
+| `detect` | ✔ | The detection block (see [Writing the detection](#writing-the-detection)). |
+| `name` | | A short internal label, for your own bookkeeping. |
+| `subject_path` | | A **top-level property key** — a single lookup, *not* a detection path — used as the finding's subject instead of the resource URN. `email`, never `event/email`. Max 256 characters. |
+| `criticality_mult` | | Risk multiplier applied to `lc_risk`, `0` to `1.6`. `0` means unset (1.0). |
+| `meta` | | Authoring metadata: `description`, `rationale`, `references`, `no_framework`, `false_positives`. Inert at evaluation time — it exists so the reasoning travels with the rule. |
+
+`resource_type` must be one of:
+
+`Account` · `AIService` · `Application` · `ComputeInstance` · `ConfigStore` ·
+`DataStore` · `DNSZone` · `EnrollmentKey` · `HasPermission` · `Identity` ·
+`Network` · `TelemetryOutput`
+
+!!! note "Only two finding classes, on purpose"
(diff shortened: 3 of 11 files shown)


lcbill (dismissed):

docs/cloud-security/api-reference.md
```diff
@@ -160,6 +160,7 @@ carry the same shape) — key fields:
 | `runtime_sids` | Sensors resolved onto the affected asset. |
 | `status`, `resolution`, `resolved_by`, `resolution_expires_at`, `owner`, `ticket` | The triage overlay. |
 | `first_seen`, `last_seen` | Lifecycle timestamps. |
+| `due_at`, `sla_source`, `sla_state` | The remediation clock. `due_at` (the deadline) and `sla_source` (the policy clause that set it, or `default`) are stored; `sla_state` — `breached` \| `due_soon` \| `on_track` \| `exempt` \| `none` — is **derived on every read** from `due_at`, `first_seen`, and `status`, because it changes with the clock and nothing writes when a deadline passes. All three are absent until the organization writes an [`sla` policy](remediation-sla.md); there is no default SLA. |
```
maximelb: `sla_state` is **not** absent without a policy. The read path stamps it unconditionally and it derives to `none` when there is no `due_at`, so every finding carries `sla_state: none` from day one — which is also what the CLI's own `--sla` help says. Only `due_at` and `sla_source` are absent.

As written the sentence also contradicts the same row's enum, which lists `none` as a value. Suggest: "`due_at` and `sla_source` are absent until the organization writes an `sla` policy; `sla_state` reads `none` until then."

docs/cloud-security/remediation-sla.md
```diff
@@ -0,0 +1,257 @@
...
+business days: business days need a per-organization holiday and region calendar
+that nothing in the platform models, and half-modelling one would be worse than
+the honest simple thing.
+
+## Stored vs derived
+
+**Stored on the finding:** `due_at` (= `first_seen` + the matched window) and
+`sla_source` (the rule name that matched, or `default`). Both are deterministic
+from the finding's first detection and the policy, so they change only when the
+finding is created or the policy is edited.
```
maximelb: This is not true when a clause matches on `owner` — which this page documents as a first-class `match` key two sections above.

The due columns are recomputed on **every** reconcile pass against the finding's *current* owner (and current severity), not only at creation and on policy edits. So assigning or reassigning an owner moves `due_at`/`sla_source` under an `owner`-scoped clause, with no policy edit involved.

That has a consequence worth stating rather than hiding: the breach latch is only carried forward when it was set against the *same* `due_at`. A deadline that moves therefore re-arms the finding and `cloud_finding.sla_breached` can fire again for it — directly relevant to the "emitted exactly once per breach" claim elsewhere in this PR.

docs/cloud-security/remediation-sla.md
```diff
@@ -0,0 +1,257 @@
...
+  that is how you say "everything else in 90 days" as the last rule.
+- The `rule` name is required and must be unique within the record. It is
+  recorded on every finding the clause dates (as `sla_source`), so an operator
+  can always see *which* clause put the date there.
+
+| `match` key | Matches |
+|---|---|
+| `severity` | The finding's severity, exact, case-insensitive. |
+| `finding_class` | The finding's class (`misconfig`, `vulnerability`, `toxic_combination`, …), exact, case-insensitive. |
+| `rule` | The detecting rule's id, exactly (e.g. `public-bucket`). |
```
maximelb: `public-bucket` is not a real rule id. The built-ins in this family are `public-data-store`, `public-data-store-sensitive` and `bucket-uniform-access-disabled`.

Since this key matches the id **exactly**, a reader who copies the example gets a clause that silently never matches — the exact failure mode the rest of this page is careful about. Use a real id.

docs/cloud-security/remediation-sla.md
```diff
@@ -0,0 +1,257 @@
...
+### Matching
+
+- **Rules are ordered and the first match wins** — the same convention the
+  [`suppression`](automation.md#suppression-rules-finding-disposition-policy) and
+  [`exclusions`](configuration.md#exclusions-the-escape-hatch) policies use. Put
+  the narrow clauses first.
+- Within a rule, keys **AND** together and lists **OR** within a key. An empty key
+  is unconstrained, and an **entirely empty `match` is a deliberate catch-all** —
+  that is how you say "everything else in 90 days" as the last rule.
+- The `rule` name is required and must be unique within the record. It is
```
maximelb: The required, record-unique field is `name`, not `rule`. `rule` is a distinct `match` key defined eight lines below (the *detecting* rule's id), so backticking `rule` here points at the wrong field.

docs/cloud-security/remediation-sla.md
```diff
@@ -0,0 +1,257 @@
...
+- An SLA is a commitment **you** make about your own remediation capacity. The
+  platform inventing one on your behalf is the same category of mistake as
+  declaring your data sensitive without being told — which is why the
+  [`classification` policy](configuration.md#classification-crown-jewels) works
+  the same way.
+- A built-in default would mean that every existing estate instantly acquires a
+  five- or six-figure `breached` count for work nobody ever promised to do. A
+  manufactured number is worse than a missing one.
+
+So adoption is a single Hive write, and the starter policy below is what to
```
maximelb, line 29: "Adoption is a single Hive write" is the only timing statement on this page, and it reads as instant. The `sla` record propagates through the same seam the custom-rules page documents explicitly: the collector composes and mirrors it on the **next collection sweep**, and the due columns are stamped on the projection after that.

Worth saying here too — an operator who writes the starter policy and immediately reads `sla_state: none` across the estate will otherwise think the record was rejected.

docs/cloud-security/findings.md
```diff
@@ -155,6 +164,20 @@ An operator's explicit disposition always wins over policy.
...
+With an [`sla` policy](remediation-sla.md) written, every covered finding carries
+a `due_at` and a derived state — `breached`, `due_soon`, `on_track`, `exempt`, or
+`none` — so the worklist can be worked by deadline as well as by risk:
+
+```bash
+limacharlie cloudsec finding list --sla breached --sort due_at
+```
+
+In the console the **Risks** table gains a sortable **Due** column and the filter
+rail a **Due** facet. There is no default SLA: with no policy record every
```
maximelb: The console labels the filter-rail facet **SLA**, not **Due** — only the table column is **Due**. Same wording at `remediation-sla.md` ("The filter rail carries a **Due** facet with a count per state").

maximelb (commented):
- `api-reference.md`: `sla_state` is never absent — it is stamped on every read and reads `none`; only `due_at`/`sla_source` are absent.
- `remediation-sla.md`: "change only when the finding is created or the policy is edited" is wrong for `owner`/`severity`-matched clauses — the due columns are recomputed every pass, and a moved deadline re-arms the breach latch.
- `remediation-sla.md`: `public-bucket` is not a real rule id, and `rule` matches exactly.
- `remediation-sla.md`: the required field is `name`, not `rule`.
- `remediation-sla.md`: the page never says the `sla` record only reaches the engine on the next sweep + projection.
- `findings.md` / `remediation-sla.md`: the console facet is labelled **SLA**, not **Due**.
- `custom-rules.md`: "countable for compliance" contradicts this page's own "custom rules do not evidence framework controls".
- `custom-rules.md`: bounds table omits the regex program-size caps (1000/rule-pattern, 5000/rule) that `matches` authors hit at write time.
- `custom-rules.md`: "no re-scan to request" reads as "live immediately" four lines above the next-sweep statement.
- Not in this diff: `docs/8-reference/cloud-security-api-iac.md` still lists the `cloudsec_policy` types without `rules`/`sla` (L65) and the `cloud_finding.*` verbs without `sla_breached` (L186-200).

docs/cloud-security/custom-rules.md
```diff
@@ -0,0 +1,493 @@
...
+
+!!! warning "Measure a broad rule before you trust it"
+    A rule that matches a quarter of its resource type is a rule that buries the
+    rest of your worklist. Before relying on a new rule, check what it actually
+    produced (`limacharlie cloudsec finding facets`) — a single near-universal
+    condition can produce six-figure finding counts on a large estate. If the
+    condition is real but the *unit* is wrong (an account-level setting reported
+    once per bucket), express it at the level it is actually fixed, or ship it at
+    `LOW`/`INFO` so it stays out of the default worklist while remaining
+    queryable and countable for compliance.
```
maximelb: "countable for compliance" contradicts this page's own **Compliance interaction** section two headings below: catalogs join controls to rule ids and no catalog names a `custom-` id, so a custom rule contributes nothing to compliance at any severity.

The LOW/INFO advice is sound as *noise* management; the compliance half of the justification only holds for built-in rules and should come out.

docs/cloud-security/custom-rules.md
```diff
@@ -0,0 +1,493 @@
...
+| Rules on any **one** resource type | 25 |
+| Overrides per organization | 1000 |
+| Composed policy size (all `rules` records together) | 512 KiB |
+| Size of one record | 256 KiB |
+| `detect` size per rule | 8 KiB |
+| `scope` operators per rule | 2, never nested |
+| `detect` nesting depth | 32 |
+| Rule id / title length | 64 / 256 characters |
+| `criticality_mult` | 0 – 1.6 |
+| Findings produced per pass | 25,000 per evaluator |
```
maximelb: The regex bounds are missing from this table, and they are write-time rejections a customer hits as soon as they use the `matches` operator this page advertises in the allowlist: **1000 compiled instructions per pattern** and **5000 per rule**, measured as RE2 *program size* rather than pattern length — a short pattern with nested groups can compile to six figures, so the number in the rejection message looks arbitrary with nothing here to explain it.

The per-organization composed `detect` budget (64 KiB, with a `scope` charged 10x its size) is likewise undocumented, and it is the bound behind the "the detect costs N units" rejection.

docs/cloud-security/custom-rules.md
```diff
@@ -0,0 +1,493 @@
...
+
+Evaluation is **stateless and per resource**. Each rule is tested against every
+resource of its `resource_type`, on every projection pass, using only that
+resource's own properties.
+
+Two consequences matter in practice:
+
+- **A new rule applies to your whole existing estate**, not just to resources
+  that change afterwards. There is nothing to backfill and no re-scan to
+  request — the next projection re-derives findings across everything already in
```
maximelb: "There is nothing to backfill and no re-scan to request" sits four lines above "The record reaches the engine at the end of the **next collection sweep**", and a reader takes the first sentence to mean the rule is live immediately.

The two statements are about different things (no finding backfill vs. policy propagation), but as ordered they read as a contradiction. Making the bullet say the rule applies to the whole estate *once the policy reaches the engine* removes it.

docs/cloud-security/api-reference.md
maximelb: Fixed in 7756bf34: the row now says `due_at` and `sla_source` are absent until a clause covers the finding, and that `sla_state` is always present and reads `none` until then.

docs/cloud-security/remediation-sla.md
maximelb: Fixed in 7756bf34: the stored/derived section now says the columns are recomputed every pass and that an owner or severity change can move the deadline with no policy edit, and the breach-events section states that the latch is held against a specific `due_at` so a moved deadline re-arms the finding.

docs/cloud-security/remediation-sla.md
maximelb: Fixed in 7756bf34: now `public-data-store`, with a note that the match is exact and a misspelled id silently matches nothing (copy it from a real finding).

docs/cloud-security/remediation-sla.md
maximelb: Fixed in 7756bf34: reworded to "Each rule's \`name\` is required and must be unique within the record".

docs/cloud-security/remediation-sla.md
maximelb: Fixed in 7756bf34: added an "It takes one sweep to land" note under the no-default-SLA section, pointing at `sync_now` to skip the wait.

docs/cloud-security/findings.md
maximelb: Fixed in 7756bf34: both pages now say the filter rail carries an **SLA** facet; only the table column is **Due**.

docs/cloud-security/custom-rules.md
maximelb: Fixed in 7756bf34: "and countable for compliance" removed; the advice now stands on noise management alone.

docs/cloud-security/custom-rules.md
maximelb: Fixed in 7756bf34: bounds table gained the 64 KiB composed detect budget (with the 10x scope charge), 1000 instructions per pattern, and 5000 summed per rule and per organization — plus a note explaining that the budget is compiled program size rather than pattern length, and what to do instead.

docs/cloud-security/custom-rules.md
maximelb: Fixed in 7756bf34: the section now opens with the sweep-then-projection seam and the `sync_now` escape, and the whole-estate bullet follows it, scoped to "the first projection after the policy lands".

lcbill (dismissed):

maximelb:
Round 1 addressed in 7756bf34 — all nine inline threads resolved. The out-of-diff finding (the Cloud Security API & IaC reference's stale policy-type and event lists) is fixed in the same commit.

All eight JSON examples were re-validated against the live hive write gate after the edits; the validation record was deleted and the test organization left as found.

docs/cloud-security/custom-rules.md
```diff
@@ -0,0 +1,505 @@
...
+a pass.
+
+A zero-violation result from a detector that did not run means nothing, so it is
+reported as what it is: not assessed. This is the same verdict the platform gives
+a control no detector covers at all, and it keeps `NOT_ASSESSED` out of the
+compliance score's denominator rather than inflating it into a green check. See
+[Compliance](compliance.md).
+
+Custom rules do not evidence framework controls: catalogs join controls to rule
+ids, and no catalog names a `custom-` id.
```
maximelb: This is wrong in one direction, and it is the round-1 finding landing on the wrong side.

Framework catalogs join controls to findings by **rule id *or* `finding_class`** (the class join is additionally scoped by the control's declared resource types). `public_exposure` is a declared class join key on controls in `cis-aws`, `cis-azure`, `hipaa`, `iso-27001`, `nist-csf`, `pci-dss` and `soc2` — e.g. the PCI DSS 1.3 control declares `finding_classes: ["public_exposure"]` with `class_scope_types: ["DataStore"]`.

So the very first example on this page (`custom-public-bucket-outside-cdn`: `DataStore` + `public_exposure`) produces findings that are attached to that control as evidence and turn it from PASS to FAIL. `misconfig` is not a class join key in any shipped catalog, so the accurate split is:

- a custom rule can never **satisfy** a control (no catalog names a `custom-` id), but
- a `public_exposure` custom rule can **fail** one.

As written, a customer who ships a noisy `public_exposure` rule will be surprised by their compliance report.

docs/cloud-security/custom-rules.md
```diff
@@ -0,0 +1,505 @@
...
+## Writing the detection
+
+`detect` is the same nested dictionary a D&R rule's `detect` block uses:
+`{"op": …, "path": …, "value": …}` leaves, composed with
+`{"op": "and" | "or", "rules": [ … ]}`, and negated by adding `"not": true` to
+any node.
+
+Paths are rooted at `event/`, and the event body is the resource's normalized
+properties. `limacharlie cloudsec resource get "lcrn:..."` prints a real
+resource, which is the fastest way to see the exact property names for a type.
```
maximelb, line 155: `resource get` cannot show two of the properties this page's own flagship example matches on.

`covers_all_ports` and `ports_effective` (used in `custom-postgres-open-to-internet` below) are **not stored properties**. They are computed onto the event at evaluation time from the raw `ports` list — a port range like `5000-20000` is expanded against a fixed sentinel-port list, and `0-65535` sets `covers_all_ports` — so they never appear in the system-of-record row `resource get` prints. The same applies to the Identity facets a custom rule may legitimately match (`dormant`, `mfa_known`, `mfa_enabled`, `credential_recently_used`, `service_linked`) and to the `HasPermission` exposure enrichment.

As written, an author who follows the stated discovery method concludes those fields do not exist — and never discovers the ones that do (which matters most for exactly the range-form firewall rules the raw `ports` match silently misses). This needs either a short per-resource-type list of the computed facets, or at minimum a sentence saying they exist and are invisible to `resource get`.

docs/cloud-security/custom-rules.md
```diff
@@ -0,0 +1,505 @@
...
+renderers the built-in rules use, so a custom finding carries a real
+offending-configuration detail. Rule-specific evidence wording and the
+ready-to-apply remediation snippets are authored per built-in rule, so a custom
+rule's findings may carry none — which is the practical reason to write the
+`title` as the action an operator must take.
+
+!!! warning "Measure a broad rule before you trust it"
+    A rule that matches a quarter of its resource type is a rule that buries the
+    rest of your worklist. Before relying on a new rule, check what it actually
+    produced (`limacharlie cloudsec finding facets`) — a single near-universal
```
maximelb: `finding facets` cannot answer "what did this rule produce". The facet response carries only `severity`, `finding_class`, `status`, `account`, `owner`, `reachable`, `in_kev` and `sla` — there is no rule dimension. For a `misconfig` custom rule the class count is shared with dozens of built-ins, so the number moves by an amount you cannot attribute.

The command that actually isolates one rule is the free-text one this page documents ~12 lines above (`finding list -q <rule-id>`), or `finding causes`. Worth pointing the warning at those instead.

maximelb (commented):
Round-2 review. Open findings:

- `custom-rules.md` — "Custom rules do not evidence framework controls" is wrong in one direction: catalogs also join by `finding_class`, and `public_exposure` (this page's own first example) is a join key on controls in 7 shipped frameworks, so such a rule can flip a control to FAIL.
- `custom-rules.md` — `resource get` is offered as the way to discover property names, but the flagship `scope` example matches `covers_all_ports` / `ports_effective`, which are computed at evaluation time and never appear on the stored row (same for the Identity facets).
- `custom-rules.md` — "check what it actually produced (`finding facets`)" doesn't work: the facet response has no rule dimension.
- `compliance.md` — "disabled every rule that evidences it" overstates the downgrade; a control that also declares `finding_classes` keeps a basis and is graded normally.

docs/cloud-security/compliance.md
```diff
@@ -38,7 +38,12 @@ The report is per-control, and each control lands in one of four states:
 - **NOT_ASSESSED** — the control cannot be evaluated automatically. These are
   the manual / attestation controls (policy documents, board sign-off,
   interview evidence): the framework marks them as not auto-assessable, so no
-  rule is run and no verdict is invented.
+  rule is run and no verdict is invented. A control also lands here when the
+  detection that would have decided it never ran — including when **you have
+  disabled every rule that evidences it** through a
```
maximelb: This overstates the downgrade. The disabled-rule path to `NOT_ASSESSED` only fires when the control's **entire** evidence basis is switched off: a control that also declares `finding_classes` keeps a basis no matter how many of its named rule ids you disable, and so does a control where one of several named rules still runs. 25 of the 69 rule-evidenced controls across the shipped catalogs are that first shape, so "you have disabled every rule that evidences it" will not produce `NOT_ASSESSED` for a substantial fraction of controls.

`custom-rules.md` states it correctly ("a control whose **only** evidence comes from rules you disabled"); this sentence drops the "only".

docs/cloud-security/custom-rules.md
maximelb: Fixed in c1fe6120: the section now says a custom rule can never make a control PASS, but that catalogs join by rule id **and** by finding class scoped to the control's resource types — so a `public_exposure` rule over a scoped type attaches as evidence and can flip a control to FAIL (the page's own first example does exactly that), while `misconfig` is not a class-join key in any shipped catalog and never moves a verdict.

docs/cloud-security/custom-rules.md
maximelb: Fixed in c1fe6120: added a note listing the facets computed at evaluation time and therefore invisible to `resource get` — the Network port facets (`covers_all_ports`, `ports_effective`, both used by the firewall example below it) and the Identity ones (`dormant`, `mfa_known`/`mfa_enabled`, `credential_recently_used`, `service_linked`).

docs/cloud-security/custom-rules.md
maximelb: Fixed in c1fe6120: the warning now says the worklist has no rule-id facet and points at `finding list -q custom-your-rule-id` instead of `finding facets`.

docs/cloud-security/compliance.md
maximelb: Fixed in c1fe6120: now reads "has disabled every rule the control relies on and it has no other basis to be graded from", matching the hedging on the custom rules page.

maximelb:
Round 2 addressed in c1fe6120 — all four inline threads resolved. The compliance-join finding was the substantive one: the round-1 wording was wrong in one direction, and the page now states that a custom rule can never make a control pass but a `public_exposure` one can fail a class-joined control.

All eight JSON examples re-validated against the live write gate after the edits; record deleted, test organization left as found.

lcbill (dismissed):

maximelb:
Round 3 (self-review, subagent budget exhausted) in 7a98368c: added the `HasPermission` computed facet to the evaluation-time list, made the per-pass finding budget legible (two evaluators, ~50k per organization) and said what blowing it costs, and stated the SLA record-name composition order plus the reject-per-record/truncate-across-records asymmetry the rules page already explained.

Verified directly against the evaluator, the collector, the hive validators, the compliance catalogs, the gateway's forwarded params and the CSV column set. All eight JSON examples re-validated against the live write gate afterwards.

lcbill (approved):

PR #370, opened 2026-09-02
docs(email-security): message pipeline and IOC feed recipe

maximelb:
Two new Email Security pages, both nav-placed and lint-clean.

## `email-security/pipeline.md` — "How a Message Is Processed"

Placed directly after **Overview**, because it is the page an evaluator wants before anything else.

Covers the twelve stages from provider notification to remediation, and is explicit about the thing product pages usually blur:

- **What is synchronous.** Fetch, parse, every enrichment *including attachment explosion*, matching, scoring, the verdict, clustering, persistence and the `EMAIL_MESSAGE` emission are one pass per message. Anything that changes a verdict afterwards is an appended revision that emits `EMAIL_VERDICT`; `EMAIL_MESSAGE` is never rewritten.
- **Managed detections are optional.** The `mailsec_policy` / `managed_rules` opt-out, the fact that the record must state `enabled` explicitly, and that with nothing matching the verdict is `unknown` rather than `benign`.
- **The state model** as six independent dimensions with one writer each, rather than one collapsed "status".
- **Storage and privacy.** AES-256-GCM, per-org keys derived from a KMS-wrapped root, storage path bound in as additional authenticated data, the 35-day transient / 400-day retained lanes, the ~256 KB per-part body cap with `body/truncated` and `_meta/truncations`, and the separately-gated EML download.
- **Latency** as three terms (provider notification delay, queue, processing) with **no invented numbers**. The page says plainly that we do not publish a latency figure yet, which matches the backend reporting that percentile as not recorded rather than estimating one.

## `email-security/ioc-feeds.md` — "IOC & Reputation Feeds"

Placed after **Custom Rules**. The recipe for mirroring a threat feed into a `lookup` with `ext-lookup-manager` and matching message links, attachments and sender domains against it.

Licensing is stated up front and not softened: fetch from the source with your own credential, keep the attribution, and read abuse.ch's fair-use terms before mirroring their data. No LimaCharlie-hosted copy of a licensed feed is referenced or recommended anywhere on the page.

### Two corrections to the original brief, worth a look

1. **The `Auth-Key` header does not work, and the page does not claim it does.** An ARL supports `basic`, `bearer`, `token` and OTX auth only; it cannot set an arbitrary HTTP header. What *does* work is that abuse.ch's v2 file exports carry the key as a **URL path segment**, so a plain `[https,host/path]` ARL is sufficient. The page documents that, plus the escape hatch (mirror to your own GitHub/GCS) for feeds that insist on a header.
2. **`op: lookup` is not usable in `dr-mail` today.** The hive validator's operator allowlist does not include it, so such a record is refused at write time, not silently ignored. The page shows the intended rule shape but marks it unambiguously as unavailable, and makes the `dr-general` rules on `EMAIL_MESSAGE` the recipe that actually works.

Other constraints surfaced because they are the ones people will actually hit: the 10 MB `lookup` record ceiling (full feed dumps exceed it), no archive unpacking, `format` accepting only `json` / `newline` / `yaml` / `optimized`, the 24h sync schedule meaning a new config is not fetched until you run a manual `sync`, and the fact that D&R lowercases the extracted value so an uppercase-hash feed matches nothing silently.

## Verification

- Every event name, path, hive name, policy type, `format` value, permission and retention number was checked against the code rather than recalled.
- Every CLI verb and flag was checked with read-only `--help` / `--ai-help`. No mutating command was run.
- `markdownlint-cli2` clean; all internal links and heading anchors resolve.
- Production was not touched.

## For the reviewer

Two claims are ahead of `master` in their own repos and should be confirmed before this merges: the `set_managed_rules` / `get_managed_rules` extension actions, and the collector-side gate on the managed pack. The `managed_rules` policy type itself is already accepted. Everything else on both pages is on `master` today or is already documented in the existing beta pages.

### DIFF
--- a/docs/email-security/index.md
+++ b/docs/email-security/index.md
@@ -99,6 +99,7 @@ Managing connections and policy uses the ordinary Hive permissions for the
 
 | | |
 |---|---|
+| [How a Message Is Processed](pipeline.md) | The pipeline end to end: what is synchronous, what is stored, and how long it takes |
 | [Getting Started](getting-started.md) | Subscribe, connect a tenant, see the first judged message |
 | [Connecting Providers](providers.md) | The connection record, credentials, scope, ingest modes and the connection test |
 | [Microsoft 365](provider-setup/microsoft-365.md) · [Google Workspace](provider-setup/google-workspace.md) | Per-provider setup |
@@ -107,6 +108,7 @@ Managing connections and policy uses the ordinary Hive permissions for the
 | [User Reports](user-reports.md) | The abuse mailbox and the report SLA queue |
 | [Detections & Verdicts](detections.md) | How a verdict is produced, and what the rules can read |
 | [Custom Rules](custom-rules.md) | Writing, validating and backtesting your own mail rules |
+| [IOC & Reputation Feeds](ioc-feeds.md) | Mirroring a threat feed into a lookup and matching messages against it |
 | [Policy Reference](policy.md) | Every `mailsec_policy` record type |
 | [Events & Automation](automation.md) | The `EMAIL_*` events and wiring them to D&R |
 | [Command Line Interface](cli.md) · [API Reference](api-reference.md) | The programmable surface |
--- a/docs/email-security/ioc-feeds.md
+++ b/docs/email-security/ioc-feeds.md
@@ -1,0 +1,333 @@
+# IOC & Reputation Feeds
+
+--8<-- "includes/email-security-beta.md"
+
+The managed rule pack carries its own link and sender reputation signals. This
+page is about the other lane: mirroring a threat feed **you** have chosen, and
+are licensed to use, into a LimaCharlie `lookup`, then matching every message's
+links, attachments and sender domains against it.
+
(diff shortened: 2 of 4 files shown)


lcbill (approved):

maximelb:
Reviewer note (correcting the PR body): the `set_managed_rules`/`get_managed_rules` extension actions and the collector-side managed-pack gate are NOT ahead of master — both merged (ext-email-security#11, legion_mailsec#48) and are deployed + live-verified on exp (ext v1.0.0-mailsec5, collector v0.1.0-mailsec41). The pipeline page's claims about them describe shipped behavior. The stale note came from reading local checkouts that were behind origin/master.

maximelb:
Reviewer note: the pipeline page's statement that the latency percentile is "not recorded" is now stale — the time-to-verdict measurement shipped and is live on exp (coverage `processing_latency_p95` returns `status: available` with `value_ms`, basis `provider_received_to_event_emitted`). Suggest updating that sentence to describe the metric before merging.

PR #377, opened 2026-09-05
email-security: document body-similarity campaign clustering

maximelb:
## Why

Campaign clustering keys on the normalized subject, the link root-domain set, and the attachment digests. A phishing kit that gives every recipient a unique subject line and a unique tracking path on every link defeats all three by construction, and sends one attack that the product reports as forty unrelated messages. Body-similarity clustering — a fuzzy hash (TLSH) of the normalized body — is the fourth key, shipping now.

These docs also carried a note saying body similarity **is not** a clustering key. That note is now wrong, and a customer reading it would conclude the product cannot do something it does.

## What changed

**`campaigns.md`** — a new **Body similarity** section:

- the fourth key added to the cluster-key table;
- what is normalized away before the body is hashed, and why each step is there (visible text only; links to scheme+host; addresses, salutation and the recipient's own name; long digit/hex/opaque tokens);
- the measured numbers rather than assertions — two copies of one 630-byte phish differing only in the recipient's name and the link's tracking parameters are **86 apart before normalization and 0 after**; across a 404-message corpus of ordinary business mail the **closest pair of unrelated messages is 39 apart**, the default join distance is **30**, and at 30 the key produces zero agreements across all 71,631 pairs;
- why it still takes **two** agreeing keys, and that the key changes how often the threshold is *reachable* rather than what it is;
- `cluster_reason`, `body_tlsh` and `body_distance` — the answer to the question a campaign-wide quarantine provokes, *why was my mail in that group?*;
- the wider **seven-day** body window and the fact that a body-similar message reopens a campaign;
- what happens to a message with no usable body (no digest, and an empty key never matches).

**`policy.md`** — the new `clustering` record: `body_similarity` (default on) and `body_similarity_distance` (default 30, range 0–35), the composition rule, why it is a separate record from `thresholds`, and the note that turning it off stops the digest counting as *agreement* without stopping it being computed or stored.

## Assumption stated

The ceiling of 35 is documented as sitting *below* the corpus's closest unrelated pair on purpose. That is the product's own justification for the bound and it is worth a customer reading it, because the obvious operator instinct on a clustering knob is to raise it.

## Public repo

Per the mailsec working agreement this is opened for review and **not merged** — it needs Maxime's sign-off.

### DIFF
--- a/docs/email-security/campaigns.md
+++ b/docs/email-security/campaigns.md
@@ -77,6 +77,16 @@ across all 71,631 pairs. The policy ceiling is 35, below that closest pair on
 purpose: a setting above it is one you cannot have measured, and what it buys is
 a campaign-wide quarantine reaching mail that was never part of the attack.
 
+!!! warning "It is closer to all-or-nothing than a tolerance"
+    At the length of ordinary email, TLSH is very sensitive: two bodies whose
+    normalized text is byte-identical score **0**, and a single per-recipient word
+    the normalization could not identify — a name in a footer, an amount, a company
+    — has been measured at **100**, well past the ceiling. So the threshold is not a
+    slider that trades recall for precision in small steps. What it buys is the mass
+    case: one pitch, randomized subjects and links. A kit that rewrites a word of
+    prose per victim will not group, and that failure is entirely on the recall
+    side — it can never merge unrelated mail.
+
 ### It still takes two keys
 
 Body similarity does **not** join a campaign on its own. It is the strongest of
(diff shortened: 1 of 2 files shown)


lcbill (dismissed):

maximelb:
Added one paragraph after live verification on exp, because the docs as written invited a reading the feature cannot support.

The exp run measured two markers whose **normalized** bodies differ by exactly one word (`sent to riley only` vs `sent to morgan only`, 622 vs 623 bytes) at **distance 100**, and a byte-identical pair at **0**. At the length of ordinary email TLSH is much closer to all-or-nothing than to a tolerance. A reader who took 30 as a slider would expect near-miss grouping that does not happen — and the natural next move is to raise the threshold, which the benign corpus says is unsafe (closest unrelated pair: 39).

So the claim is now bounded rather than removed: what the key buys is the mass case — one pitch, randomized subjects and links, which is the shape it was built for and which is live-proven — and its failure mode is recall only, never a merge of unrelated mail. Tracked on the mailsec board for a measurement against a real multi-recipient campaign.

/lc-review

lcbill (approved):

PR #389, opened 2026-09-08
docs(email-security): the read budget, and why free text needs a window [HOLD until the gateway ships to prod]

maximelb:
Documents the behaviour landing in **lc_api-go#944** (security review **L-12**), board ticket *"[Pre-GA] No read-cost governor on the mailsec read surface"*.

Two reads on the Email Security surface recompute rather than serving a cached or seekable answer, and both are now bounded per organization: `GET /coverage` over an explicit window, and `GET /messages` carrying `q`. Neither was documented as costing anything, and the filter table's *"Free text over the message's identifying fields"* gave no hint that `q` is matched **row by row** while every other filter in that table is a **lookup**.

- **api-reference.md** — a `## Read budgets` section: which two reads, the 2,400/hour figure, the `429` body, `rate_bucket: mailsec_read` as what distinguishes it from the raw-download refusal on the same surface, `Retry-After`, and the two cheapest ways to get a refused read served. The `GET /coverage` and `GET /messages` rows now point at it.
- **messages.md** — `### Free text needs a window`: what `q` actually costs, exactly what counts as a bound (`since`, the five lookup filters, a **single** verdict) and what does not (`until` alone, and the residuals), and the 512-character cap.
- **cli.md** — `coverage` with no window is memoized and is the right shape for a poller; `--window-days` recomputes and is budgeted.
- **automation.md** — the same note where the docs already recommend scripted coverage polling.

The `q` examples are HTTP rather than CLI on purpose: `q` has no CLI flag today (a parity gap filed separately).

Anchors checked by hand: `api-reference.md#read-budgets` (new), `#free-text-needs-a-window` (new, depth 3 so within `toc_depth`), and `#get-messagesmsg_uuideml`, which detections.md already links to.

**Public repo — not merging this myself.** Ready for review whenever it suits.


lcbill (dismissed):

lcbill (dismissed):

lcbill (dismissed):

maximelb:
**HOLD — merge with the gateway prod release.**

This documents behaviour that does not exist on `api.limacharlie.io` yet. The rule it describes (a bare `q` is refused; the 512-character cap; the read budget) ships in lc_api-go#944, which is going to **legion-exp only** for now. The docs site publishes on merge, so merging this before the gateway reaches production would tell customers their working `?q=…` calls are refused while they still succeed — and the `429` shape and `rate_bucket` would document a budget that is not being charged.

Reviewed and green (markdownlint, link check, rendered build). Please merge it in the same wave as the lc_api-go production release, not before.

lcbill (dismissed):

lcbill (dismissed):

lcbill (approved):

PR #390, opened 2026-09-08
docs(email-security): body similarity groups by normalization, not by tolerance

maximelb:
## Why

The campaigns page's Body similarity section made two claims from single measurements and
they read as guarantees:

- *"86 apart before normalization and 0 apart after it"* — real, but one synthetic pair;
- a warning that a single unremovable per-recipient word *"has been measured at 100"* — also
  real, and also one pair.

Re-measured across a whole corpus rather than on one pair, the cost of one per-copy word the
normalization cannot identify has a **median of 20–30 and a p95 over 100** (worst 201). So the
old text overstates the failure in the typical case and understates the variance, and — the
part that matters operationally — it leaves "raise the threshold" looking like a reasonable
response. It is not: moving 30 to the ceiling of 35 takes that case from roughly half to
roughly two thirds while spending most of the margin against unrelated mail.

Meanwhile the product side of this shipped five more normalizers (all recipients' names, wider
greetings, signature blocks, quoted-reply tails, short mixed letter-and-digit references, and
whole-number collapse), so the list of what the normalization removes was out of date and the
benign-corpus margin moved 39 → 40.

## What changed

- The normalization list now describes what is actually removed.
- The measured example is now a **kit templated over eight recipients in three shapes**
  (individually addressed, shared mailbox, thread-hijack) — 37–219 apart raw, 0 apart
  normalized — rather than one pair.
- The warning admonition is rewritten so the threshold is not read as a tolerance dial, and
  says plainly that raising it is not the lever it looks like.
- 39 → 40 for the closest unrelated pair, in both places it appears.

No behaviour change here; this is the public wording catching up with the measurement.

## Hold

**Do not merge yet** — this describes normalizers that are not in production. Merge with the
prod release, same as #388 / #389.

### DIFF
--- a/docs/email-security/campaigns.md
+++ b/docs/email-security/campaigns.md
@@ -90,9 +90,9 @@ the body is reduced to what it actually says:
 
 How much this matters, measured rather than claimed: one phishing pitch templated
 over eight recipients — name, greeting, amount, account fragment, tracking token
-and signature all varying per copy — is **39 to 135 apart before normalization and
-0 apart after it**, in all three shapes such a kit takes (addressed to each victim,
-collected from a shared mailbox, or top-posted above a stolen thread). The default
+and signature all varying per copy — is **37 to 219 apart before normalization and
+0 apart after it**, across all three shapes such a kit takes (addressed to each
+victim, collected from a shared mailbox, or top-posted above a stolen thread). The default
 join distance is 30, so without the normalization this key would not work at the
 length of an ordinary email.
 
--- a/docs/email-security/campaigns.md
+++ b/docs/email-security/campaigns.md
@@ -101,9 +101,9 @@ length of an ordinary email.
 Two bodies count as the same body at a **distance of 30 or less**, which is a
 policy knob (`clustering` — see the [Policy Reference](policy.md#clustering)).
 
-30 is measured. Across a 404-message corpus of ordinary business mail —
-newsletters, invoices, calendar invites, internal notices — the **closest pair of
-unrelated messages is 40 apart**, and at 30 the body key produces zero agreements
+30 is measured. Across a corpus of several hundred pieces of ordinary business
+mail — newsletters, invoices, calendar invites, internal notices — the **closest
+pair of unrelated messages is 40 apart**, and at 30 the body key produces zero agreements
 across every pair. The policy ceiling is 35, below that closest pair on purpose: a
 setting above it is one you cannot have measured, and what it buys is a
 campaign-wide quarantine reaching mail that was never part of the attack. That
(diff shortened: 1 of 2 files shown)


lcbill (dismissed):

maximelb:
**HOLD — do not merge yet.** Approved and green, but this describes normalizers that are not in production: it depends on go-mailsec#134 and its legion_mailsec re-pin reaching the prod datacenters. **Merge with the prod release**, alongside #388 and #389.

lcbill (dismissed):

lcbill (approved):

maximelb:
**Held — several of the revised numbers and one normalization step do not match the code.**

The dependency this was waiting on has shipped (`go-mailsec#134` is in `v0.1.76`, which `legion_mailsec v0.1.0-mailsec131` pins, running in all six production regions), so the hold is no longer about sequencing. It is about accuracy. I ran the measurement harness in `bodyhash/` against `origin/master`; here is what it says.

**Confirmed, keep as written:**

- Closest pair of unrelated messages is **40** (`payroll-payslip-notification` / `payslip-available-portal-notice`). The old 39 is genuinely stale.
- **Zero** agreements at distance 30 across every pair (86,736 of them), and zero at the ceiling of 35 too.
- Default 30, ceiling 35, the `clustering` knob, byte-identical normalized text scoring 0, the two-key rule and the four keys.
- Median cost of one unidentified per-copy word is **29**, p95 is **124** — so "median of 20–30" and "exceeds 100 in the worst 5%" both hold.

**Wrong or misleading:**

1. **"URLs reduced to registrable domain."** `urlStem` keeps scheme + the *entire* authority, subdomains included; it drops path, query and fragment and rewrites userinfo `@` to ` at `. Registrable-domain reduction is the *link* cluster key — a different thing, and conflating them in a page that explains both is worse than saying nothing.
2. **The step order.** The list has numbers collapsing before hex/opaque/short tokens. The code is the reverse and says why: the id rule reads digits, so collapsing them first hides every mixed-alphanumeric reference. Quoted-thread collapse also runs *before* address collapse (an attribution line is identified by containing a real address), and signature collapse runs before the greeting.
3. **Two steps are missing**, and one of them is load-bearing: lowercasing, stripping Unicode format/control runes, and — the important one — every token rule runs *masked outside link authorities*. That mask exists because of a measured false-merge defect; a reader tuning against this list would not know it is there.
4. **"Roughly half to roughly two thirds"** measures as **56% → 71%**. The direction of the argument survives; the numbers should say what they are.
5. **The 37–219 range is real but not reproducible from a shipped test.** `TestMeasureMultiVictimKit` only measures raw distances for the first of the three kit shapes (kitA: 39–135). I had to add the computation for kitB and kitC to get 37 and 219 (all 84 raw pairs; all 84 normalize to 0). If the doc is going to cite the number, the test should print it — otherwise the next person to touch the normalizers has no way to re-derive it.

**Unrelated, but found while checking:** `go-mailsec` `policy/types.go:851` still tells operators the closest benign pair is 39 when it is 40. Worth a one-line fix in that repo.

Everything else here is an improvement on what is published. Leaving it open for the corrections rather than merging measured claims that do not reproduce.

PR #394, opened 2026-09-08
docs(access): resource ACLs — restricting sensors and configs by scope

maximelb:
User-facing documentation for resource ACLs. New page at `docs/7-administration/access/resource-acls.md`, plus a nav entry and three small edits to existing pages.

## The page

Written for an administrator who wants to compartmentalise part of an organization — a mail feed with message bodies, an HR fleet, a customer covered by a separate agreement — and has to decide whether to do that with a separate org, a group, or a resource ACL.

- The model: an `acl:<scope>` tag on a resource, an `acl` hive record listing the scope's members, and the rule that both the ordinary permission check and scope membership must pass.
- A table of what is gated (telemetry, tasking, detections, artifacts, config record contents) against what stays visible (existence, hostname, tags, online status, counts), so nobody expects sensors to disappear from the fleet list.
- `acl.set` / `acl.get`, which roles carry them, and a plain statement that `acl.set` is confidentiality-equivalent to unrestricted read while granting no *implicit* read.
- Step-by-step setup with the CLI: create the scope, choose member types, tag sensors / installation keys / config records, verify — including "check with a real analyst account, not one holding `acl.set`".
- Outputs: excluded by default, opt in per scope with `acl_scopes`, live streams inherit the opener's scopes, and the retention-output caveat.
- D&R rule `acl_scopes`, the redaction marker on config records and why writing it back is refused, and the extension-API-key gotcha for restricted playbooks and configs.
- Error-code table and a troubleshooting section.

## The behaviours that surprise people

These get their own section because every one of them is counter-intuitive and at least one is a footgun:

- Restriction follows the tags a resource carries **now**, so tagging a sensor hides its history and untagging re-exposes it.
- A scope tag whose record does not exist, is disabled, or has expired **locks** the resource. Deleting a scope record is not how you un-restrict something — removing the tag is.
- Scope tags cannot carry a TTL, and D&R rules cannot add or remove them.
- Propagation is bounded at about five minutes, so removing someone from a scope is not instant revocation.

## Other changes

- `acl.set` / `acl.get` added to the permissions reference.
- A short "reserved `acl:` namespace" section in Sensor Tags, since that page currently tells people tags are free-form.
- A pointer from Designing Access, whose "groups are additive only" warning previously left "how do I take access away" unanswered.

## Verification

- Every CLI command, flag and hive name was checked against the installed `limacharlie` CLI (5.6.3), not from memory. The `acl_scopes` output parameter was traced through the SDK's form encoding to confirm a YAML list in `--input-file` arrives as a list.
- `markdownlint-cli2` clean; all relative links resolve.
- No screenshots, since the flows documented here are CLI and API.

### DIFF
--- a/docs/2-sensors-deployment/sensor-tags.md
+++ b/docs/2-sensors-deployment/sensor-tags.md
@@ -40,6 +40,17 @@ You can use tags to create workflows and automations. For instance, you can conf
 
 Create a Yara scanning rule so that endpoints tagged as 'sales' are continuously scanned against the specific sets of Yara signatures.
 
+## The reserved `acl:` namespace
+
+Tags beginning with `acl:` are reserved. They restrict who may read a sensor's
+telemetry, task it, or see its detections, and adding or removing one requires
+the `acl.set` permission on top of the usual `sensor.tag`. They also cannot be
+given a TTL, and Detection & Response rules are not permitted to add or remove
+them.
+
+Use any other tag freely — the reservation applies only to the `acl:` prefix.
+See [Resource ACLs](../7-administration/access/resource-acls.md).
+
 ## Adding Tags
 
 Tags can be added to a sensor a few different ways:
--- a/docs/7-administration/access/designing-access.md
+++ b/docs/7-administration/access/designing-access.md
@@ -114,6 +114,8 @@ End customers must stay confined to their own organization. The safe pattern is 
 !!! warning "Groups are additive only"
     Permissions granted through a group are **added** to the user's direct permissions on each included organization. A group cannot be used to *reduce* or *restrict* what a user can see. Treat "membership in a group" as "give every permission in that group, on every org in that group."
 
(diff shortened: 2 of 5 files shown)


lcbill (dismissed):

lcbill (dismissed):

lcbill (dismissed):

lcbill (dismissed):

lcbill (approved):

maximelb:
Superseded by #427, which documents the same feature in a shorter form and has been checked line by line against the current backend code.

Everything material from this branch was folded into #427: export-to-output dropping all restricted records, retention outputs, live-stream scopes and the `acl.set` non-bypass, the public/unauthenticated endpoint refusing tagged records (adapters pinned to last known-good config), the tagged-sensor prerequisite for investigation expansion and extension envelopes, the `INVALID` and `error_code` details, audit/billing notes, the `dr set --detect/--respond` trap, and the Designing Access cross-link.

Differences left deliberately: #427 does not document `sync --hive-acl` as available today (the flag is merged but not in the latest CLI release, 5.6.2), and it drops the worked example and some of the longer explanatory sections to keep the page short.

Closing this one; reopen if you prefer the longer treatment.

PR #399, opened 2026-09-09
email security: document the replay budget on POST /rules/backtest and POST /hunts

maximelb:
Documents the second Email Security read budget added by **lc_api-go#946** — the one covering `POST /rules/backtest` and `POST /hunts`.

**Stacked on #389 deliberately.** That PR introduces the `## Read budgets` section for the two GET reads; this extends it rather than opening a competing section, so the base here is `mailsec-read-governor-docs` and not `master`. Merge #389 first.

## What it says

- The section now has **two** subsections — *The query budget* (unchanged, #389's) and *The replay budget* — with an opening paragraph saying why there are two: the reads are expensive in different resources, are counted separately, refill on different clocks, and carry different `rate_bucket` values.
- The replay budget in the terms a customer can act on: it is counted in **messages re-read**, one call is charged the most it could re-read (2,000), the allowance is **12,000 per 10 minutes = 6 backtests per 10 minutes** across every credential in the org, and it **fails open**.
- `rate_bucket: mailsec_post_read` is **not** `mailsec_read`, stated explicitly with the consequence: a client that treated them as one budget would retry a backtest on advice that does not apply to it.
- The honest bit the query budget's advice does not carry over: **no narrower request shape takes a backtest out of this budget** — the charge is the same whatever window you name. A filtered search leaves the query budget entirely; a narrower backtest does not. Asking for a narrower window is still worth doing, and the text says why (speed, not budget).
- Why `POST /hunts` is counted while it still answers `not_implemented`: budgeting it on the day the engine lands would be a change of contract for clients that had already shipped.
- Notes on the two pages a rule author actually reads — `custom-rules.md` (under Backtesting) and `cli.md` (beside `rule backtest`).

## Held

**This must not be merged until the gateway reaches production.** The docs site publishes on merge, and until lc_api-go#946 ships to `api.limacharlie.io` this page would tell customers their backtests are budgeted while they are not.

Anchors were checked by hand — this repository does not validate URL fragments. `#the-replay-budget` and `#registered-but-not-implemented-yet` both resolve, and the existing `#read-budgets` links from `automation.md`, `cli.md` and `messages.md` still resolve because the H2 is unchanged.

### DIFF
--- a/docs/email-security/api-reference.md
+++ b/docs/email-security/api-reference.md
@@ -186,8 +186,18 @@ organization is deleted. See
 
 ## Read budgets
 
-Two reads on this surface recompute something rather than serving a cached or
-seekable answer, and both are bounded per organization.
+Some reads on this surface recompute something rather than serving a cached or
+seekable answer, and those are bounded per organization. There are **two
+separate budgets**, because the reads they cover are expensive in different
+resources: the two below are measured in database work, and the
+[rule replays](#the-replay-budget) further down are measured in **stored
+messages re-read**. They are counted separately, refill on different clocks, and
+carry different `rate_bucket` values — so do not retry one on the other's
+advice.
+
+### The query budget
+
+Two reads recompute a query rather than serving a cached or seekable answer.
 
 - **`GET /coverage` with an explicit window** — a `since`/`until` pair, or a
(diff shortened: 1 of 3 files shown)


maximelb:
merge with the prod release

lcbill (approved):

PR #400, opened 2026-09-09
docs(email-security): body/current_thread/text covers both renderings of a message

maximelb:
Documents a contract change landing in go-mailsec (private PR, build half of a MailSec board ticket): `body/current_thread/text` — the path this site's custom-rules page tells a customer to write against — now carries the newest segment of **both** of a `multipart/alternative` message's renderings rather than one of them chosen by a length heuristic.

**Why the change exists.** The two renderings of a message are not required to agree, and a sender who padded the `text/plain` alternative past a fixed byte bound could put the real message in the `text/html` part a reader sees and have every rule matching this path read the padding instead. The fix removes the choice: the field is the union, with the HTML half omitted only when it already appears word for word in the plain part.

**What a rule author needs to know**, and what this note says:

- Match with substring/regex patterns rather than whole-value equality — on a message whose renderings differ, the field says the same thing twice.
- It is a matching surface, not a display one: render `body/html/display_text` or `body/plain/raw` to show a person the message.
- The quoted history is still excluded, and still available under `body/previous_threads`.

**HOLD — do not merge.** This site publishes on merge and the statement is only true once the change reaches production; production is on hold.

### DIFF
--- a/docs/email-security/detections.md
+++ b/docs/email-security/detections.md
@@ -230,6 +230,19 @@ chain).
     The quoted history is still excluded from it — that is the point of
     `current_thread` — and it is available separately under `body/previous_threads`.
 
+!!! warning "A clause that switches a rule OFF must read `body/current_thread/visible_text`"
+    Covering both renderings is the right answer for a clause that looks for
+    something, and the wrong one for a clause that calls a rule off. A rule of the
+    shape "fire on X **unless** the message also says Y" reading `text` can be
+    switched off by a sender who writes X into the part you read and Y into the
+    part you do not — a suppressor the recipient is never shown.
+
+    `body/current_thread/visible_text` is the same newest segment narrowed to the
+    rendering a reader is actually shown: the text extracted from the HTML when the
+    message has an HTML part, the plain part when it does not. Write positive
+    clauses against `body/current_thread/text` and any `not: true` clause against
+    `body/current_thread/visible_text`.
+
 ### Enrichments
 
 | Path | What it carries |
--- a/docs/email-security/detections.md
+++ b/docs/email-security/detections.md
@@ -210,6 +210,26 @@ quoted threads, hidden-text detection), `links`, `attachments`, `auth` (parsed
 SPF / DKIM / DMARC / ARC results with alignment) and `hops` (the parsed `Received`
 chain).
 
+!!! note "`body/current_thread/text` covers every rendering of the message"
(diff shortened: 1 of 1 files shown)


maximelb:
merge with the prod release

lcbill (dismissed):

maximelb:
Updated: the note now also tells a rule author to point any `not: true` clause at `body/current_thread/visible_text`, which is the same newest segment narrowed to the rendering a reader is actually shown. Still HELD — merge with the prod release.

lcbill (approved):

maximelb:
**Held — the warning block is now describing behaviour the product has moved past.**

The `!!! note` half of this PR is correct and verified against `go-mailsec` `parser/parser.go`: `body/current_thread/text` is the union of both renderings, plain first, joined by a blank line, with the HTML segment omitted only on exact containment; `visible_text` is the HTML display text when there is an HTML part and the plain part otherwise; quoted history stays in `body/previous_threads`.

The `!!! warning` half is the problem. It tells a rule author that a `not: true` clause **must** read `body/current_thread/visible_text`, because a suppressor read off the union can be satisfied by text the reader never sees. That was true when this was written. It is not true of what is running now: `go-mailsec#229` ("signals: evaluate current-thread rules per rendering") changed the engine so that any rule touching a `body/current_thread/*` path is evaluated **once per rendering**, with `text` and `visible_text` both narrowed to that rendering, and the results OR'd (`signals/signals.go`, `matchCompiled` / `renderingMessages`). A suppressor hidden in the part the victim does not read can therefore only stand down the rendering it is in — which is precisely the attack this block tells people to defend against by hand.

That commit is in `go-mailsec v0.1.76`, which `legion_mailsec v0.1.0-mailsec131` pins, and that image is running in all six production regions. It applies to the org's own `dr-mail` rules too — they go through the same `signals.Pack.Match`.

Two consequences for the rewrite:

- The advice should be that the engine handles this, not that the author must. `visible_text` is still the right path when you specifically want "what the reader was shown" as a fact in its own right, which is a different and narrower claim.
- "On a message whose two renderings differ, this field says the same thing twice" is true of the stored MDM and of the `EMAIL_MESSAGE` event, but **not** of what a rule sees — a rule sees one rendering at a time. Worth separating, since the note is addressed to rule authors.

Everything else in the PR can stand. Leaving this open rather than merging a page that would have customers write rules around a problem the engine already solves.

PR #402, opened 2026-09-09
docs(email-security): the message search is literal, and blank text is not a filter

maximelb:
Publishes the contract that go-mailsec#163 and lc_api-go#947 implement.

**What was wrong.** The API reference lists `q` among the message-index filters and says
nothing about how the text is matched. It was matched as a SQL pattern, so `%` and `_` were
wildcards and a backslash was an escape character: a search for `50% off` also returned
`50 percent off`, a search for `invoice_2026` also returned `invoiceX2026`, a lone `%`
returned every message in the index, and `c:\temp` returned the row *without* the backslash
and missed the one with it. None of that was documented, taught or refused — it was an
accident of the query, and the fix is to match literally rather than to document wildcards
nobody asked for.

**What this adds.** One bullet under "Shared behaviours": `q` is a literal, case-insensitive
substring over the message's subject and sender address; `%` and `_` are ordinary characters;
surrounding whitespace is ignored and whitespace-only text is not a filter.

It is placed in the shared-behaviours list rather than in the `GET /messages` row on purpose:
the held read-budget PR (#389) rewrites that row, and two PRs editing the same line would
collide at merge time. The two statements compose — that one is about what a search must be
accompanied by, this one about what the text means.

**Hold.** This describes behaviour that is not in production until the Email Security gateway
and collector ship there, and Email Security's prod rollout is on hold. Do not merge until
the prod release.

### DIFF
--- a/docs/email-security/api-reference.md
+++ b/docs/email-security/api-reference.md
@@ -31,6 +31,11 @@ Shared behaviours:
   `?verdict=malicious&verdict=suspicious`. OR within a key, AND across keys.
 - **Boolean selectors are tri-state.** An absent parameter means "not filtered",
   which is *not* the same as passing `false`.
+- **Free-text search (`q`) is literal.** It is a case-insensitive substring match
+  over the message's subject and sender address. `%` and `_` are ordinary
+  characters rather than wildcards — searching for `50% off` returns mail whose
+  subject contains `50% off`, and nothing else. Surrounding whitespace is
+  ignored, and a search that is only whitespace is not a filter at all.
 - **Keyset pagination.** Pages carry `next_cursor`; pass it back as `?cursor=`.
   An empty `next_cursor` is the last page. A cursor is **bound to the filter set
   that minted it** — changing a filter mid-walk fails the next page rather than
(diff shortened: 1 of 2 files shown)


maximelb:
merge with the prod release

lcbill (dismissed):

lcbill (dismissed):

lcbill (approved):

maximelb:
merge with the prod release

maximelb:
Closing — superseded by #389, which is merged.

Both PRs documented the same two things (`q` is literal; some reads are budgeted). #389 matches the code and this one is a release behind it in two ways:

- **It omits the second budget.** There are two, on different clocks and with different `rate_bucket` values: `mailsec_read` (7,200/hour, `coverage_window` and `message_search`) and `mailsec_post_read` (6 backtests per 10 minutes, `rule_backtest` and `hunt_create`). A client told there is one budget would retry a backtest on the other one's `Retry-After`.
- **Its rule for when a `q` is counted is wrong.** "Counted only when it carries a `q` that nothing but time bounds" — a `q` bounded by a *single `verdict`* is permitted by the narrowing check and is **still charged**, because the verdict index is keyed by verdict then time and seeks into what is, for most organizations, all of their mail. Only `mailbox` / `sender_email` / `campaign_id` / `link_domain` / `attachment_sha256` exempt it.

Everything this PR got right is in #389, plus two corrections that came out of verifying it against the code: `q` matches the subject in **both** its raw and normalized forms (not just the displayed subject), and the EML download's fail-closed `503` carries only `error` — `refused_reason: quota_unavailable` goes to the audit trail, not the response body.

Nothing to salvage here.

PR #408, opened 2026-09-12
email-security: engine_version names the rule pack and the library build it ran with

maximelb:
Wording for `engine_version`: it now carries the rule-pack version plus the build of the parsing and enrichment library the rules ran with, so two verdicts with the same value came from the same engine. Three lines across detections.md and automation.md. Held for the production release.


maximelb:
merge with the prod release

lcbill (approved):

PR #414, opened 2026-09-15
Cloud Security: when a GitLab or Bitbucket token is refused, and code scanning being GitLab.com only

maximelb:
Follows the GitLab and Bitbucket Cloud connector guides with three things an operator hits that the pages did not say.

**The broad-token refusal is not only a connection test.** Neither provider can narrow a token per repository, so each scan clones with the connection's *own* token. The refusal of `api` / `admin_mode` / `sudo` (GitLab) and `admin:` / `delete:` (Bitbucket) is now applied when a scan asks for the credential and on every sweep, so a secret rotated to a broader token *after* the connection was created is refused too. Both Verify tables say so, and both Troubleshooting tables carry the error an operator will actually see.

**Bitbucket: a token created without scopes.** The API then reports no scopes, and that reads as a benign "not verified" today. It is not benign — such a token is bounded by nothing and carries the full reach of the account that made it, so it is refused. Written down as a cause with its fix.

**GitLab: code scanning is GitLab.com only.** A self-managed instance connects, and its projects are inventoried and assessed exactly as before. Repository cloning runs in an isolated, egress-restricted environment that does not reach customer-run instances, so scans of those projects cannot complete. The connection test now reports this up front as `code_scanning_reachable`, and the page carries the symptom — "the inventory appears but scans never produce findings" — in Troubleshooting.

Docs only; no navigation or structural changes.


lcbill (dismissed):

lcbill (dismissed):

lcbill (dismissed):

maximelb:
### Adversarial review (self-run — `/lc-review` produced no AI review body on this repo)

Two rounds against the diff **and** the implementation in the connectors, since every claim here is a claim about code.

**Round 1 — two findings, both mine, both fixed in `15eff82c`:**

1. **HIGH — I overstated my own change.** The Bitbucket limitation said a token created *without* scopes is refused "at scan time **and on each sweep**". Only half true, and the difference is deliberate: an `admin:`/`delete:` scope is refused by the connection test, by code scanning *and* by the inventory sweep, while an unscoped token is refused by the test and by code scanning only — the sweep never hands the token to anything, so the estate keeps refreshing. Now written as two distinct conditions, with the `token_scopes` row saying what an absent scope list actually means.
2. **MEDIUM — a pre-existing gap.** `namespace_membership` is a REQUIRED GitLab check, re-checked on every sweep, and appeared in **neither** GitLab table — while the Bitbucket page documents its equivalent in full. Added to both, with why it exists (a public or internal group answers a non-member with its **public** projects only, so an unconfirmed listing cannot be trusted as the estate) and the fix (Reporter or above; a group access token's bot user is a member by construction, a personal token's owner is not).

**Round 2 — one finding, the markdown lint, fixed in `303b7824`.** `MD007/ul-indent` wants a 2-space sub-list indent and python-markdown wants 4 to nest at all; rather than pick, the two clauses are prose in the same bullet. `check-markdown` now passes.

**Verified against the code, no issues:** all 15 check IDs with their exact required/advisory flags and orderings; both quoted error strings verbatim against each provider's `ErrTokenTooBroad`; the "code scanning is GitLab.com only" claim and its bounds against `selfManaged()` and the egress rationale; public-repo hygiene (no private repo names, internal hostnames, customer names or internal jargon).

The connector behaviour these pages describe ships in refractionPOINT/legion_cloudsec_host#247.

lcbill (dismissed):

lcbill (dismissed):

maximelb:
**Updated for the rule change, plus two adversarial rounds** (`24ee0288`, `eacf5f80`, `160940c0`). The connectors no longer refuse an over-broad token: they accept any token carrying at least the required scopes and refuse only a **missing** one.

- Both check tables lose the `token_not_tenant_wide` / `token_not_workspace_admin` rows — those checks were deleted, not downgraded — and both warning admonitions now say a broad token is **accepted**, while still stating plainly what is being accepted (each scan clones with the connection's own token, into a job that analyses untrusted third-party source). A Bitbucket token created without scopes is reported, not refused.
- What stayed true is now stated more precisely, since it is the whole of the enforcement: a **missing** required scope is refused where the token is **used**, not only in the connection test, and a sweep refusal leaves the inventory **unchanged** rather than reporting an empty estate.

Two adversarial rounds against the diff **and** the Go implementation, from a pristine `git archive` extraction (this repo produces no AI review body):

- **Round 1, three HIGH:** the reverted rule was still stated on `providers.md` (both providers) and `code-scanning.md` — pages this PR had never touched, so a diff-only review would have missed all three. Plus two pre-existing corrections: GitLab `token_expiry` is advisory in the test but the *minter* refuses a revoked token before every scan; Bitbucket `token_read_only` is not emitted at all when the API reports no scopes.
- **Round 2, one MEDIUM:** both pages said a token "missing one of the required scopes" is refused, but the code tests an any-of **capability** set — a GitLab token holding only `api` is literally missing both named scopes and yet connects, sweeps and scans, and that is precisely the token a customer arrives with. Reworded on both pages and in all four table rows. Two LOWs also taken: the GitLab admonition's absolute claim needed the self-managed no-introspection exception, and `read:user:bitbucket` has no scope check anywhere — it is enforced only by `GET /2.0/user` failing.

Round 2 independently re-verified all 18 check IDs and their `required` flags against `report.AddCheck`, both quoted error strings against the Go source, every "where it runs" claim against which file the check lives in, that every call in both packages is a `GET`, public-repo hygiene, and — by a whole-tree grep — that no "broad token is refused" claim survives anywhere in `docs/`.

markdownlint clean. Public repo, so still yours to merge.

lcbill (approved):

PR #416, opened 2026-09-16
Code scanning: how a pull request actually triggers a check

maximelb:
## Summary

Turning `pr_checks: true` on and granting the App permissions does **not** make a check appear on a pull request. A `pull_request` webhook and a relay rule are what tell Cloud Security that a pull request happened, and this page described that second recipe rule in a single paragraph without ever giving it. A reader could complete every documented step and end up with a feature that never fires and raises no error anywhere.

Three commits, in `docs/cloud-security/code-scanning.md`:

**1. The rule, in full.** A new **Fire the check from the webhook** section with the YAML, and the fact that `pull_request` goes on the **same** hook as `push` rather than a second one. A warning callout early in the pull-request section, because this is the step most easily missed. The feature table gains a **Trigger** column.

**2. `pr` is a path, not a template.** A D&R `extension request` value written as `{{ ... }}` is rendered to text; a bare path keeps the payload's own type, and the pull-request *number* has to arrive as a number. Verified live: with the templated form the rule fires and the request is rejected before Cloud Security sees it — `invalid value for pr: not an integer, a string`, four times on the redelivery loop, with nothing on the pull request. The YAML says so inline, a warning callout explains why one field is written unlike the rest and what the failure looks like, and **Is it firing?** gains the step between "the rule matched" and "the check ran".

**3. The install instructions describe something that exists.** The page said the Cloud Security **Code** page "offers to write" each recipe rule. It does not — nothing in the console writes either record. Both steps now give the `limacharlie hive set` command, which is what the rest of the page already uses for hive records. This was true of the push rule too, so both are fixed.

Also: **draft pull requests are checked**, stated explicitly with the reason — the service reacts to `opened`, `synchronize` and `reopened` and **not** to `ready_for_review`, so a rule that skipped drafts would leave a draft marked ready with no further push with no check at all, and a required check would then block that merge forever. The one-condition fork for teams that do want drafts excluded is given, with the trade-off spelled out.

And a new **Is it firing?** walk-through, in order: the GitHub delivery (and what a `401` means), a rule replay, whether the request was accepted, the pull request's Checks tab, and what each refusal reason means (`pr_checks_disabled`, `write_app_not_configured`, the two write-budget ones).

## Coordination

**Stacked on `appsecga77-code-webhook-hmac`** (PR #413, the signature-verification docs), which edits the same page. Nothing was pushed to that branch. Merge #413 first and GitHub retargets this one to `master`.

## Test plan

- [x] `markdownlint-cli2` over `docs/**/*.md`: `Summary: 0 issues in 0 files`. CI green on the final head (`Docs Preview`, `Link Checker`, `Snippet Compilation`, `Python Tests`).
- [x] Every CLI flag used here was checked against the CLI's own `--help` rather than written from memory: `replay run --name/--start/--end` (an earlier draft used `--rule-name --rule-namespace --events`, which do not exist) and `hive set --hive-name/--key/--input-file/--enabled`.
- [x] The YAML matches the rule the product ships, field for field, including the `pr` path.
- [x] Proven live end to end before this was written: `opened` and `synchronize` each produced a check run on the pull request's own head commit; `review_requested` and `closed` produced nothing.

---
Reopened from #415, which GitHub auto-closed when its stacked base branch was deleted on merge. Same commits, rebased onto master; #415 was green on all 7 checks.


lcbill (approved):

maximelb:
Heads-up from the #106 session (board ticket `maximelb/claude-config#106`): **#417 is stacked on this branch** — it replaces this page's "retargeting the base branch does not re-run the check" note with the rule that does.

Please **do not merge with `--delete-branch`** while it still targets `appsecga90-pr-relay`: deleting the base auto-closes the child and GitHub then refuses to reopen or retarget it. Either retarget #417 to `master` first, or merge without deleting and I will rebase.

Same arrangement in web-app-frontend: #4391 is stacked on #4390.

PR #417, opened 2026-09-16
Code scanning: the check follows a pull request that changes its base branch

maximelb:
> **Stacked on #416** (`appsecga90-pr-relay`), whose page this extends. Retarget this PR to `master` **before** merging #416 — merging a base branch with `--delete-branch` while a child targets it auto-closes the child.

## What changes

The page currently tells a reader that retargeting a pull request's base branch does **not** re-run the check, and offers a workaround (push a commit, or close and reopen). That note is replaced by a rule that does it.

Changing the base changes what a pull request introduces without pushing anything — the provider reports it as an `edited` action rather than a `synchronize` — so the check already sitting on the head commit stayed exactly as it was, measured against a base that is no longer the base. For a required check, that is a merge gate satisfied by a scan of a diff that no longer exists.

## What the new section says, and the two traps it names

- **Why it is a second rule.** A rule's list of actions is an *or*, so one rule cannot say "`edited` **and** the base moved" — and adding `edited` to the existing rule would re-scan on every title edit: two scans and two writes into the reader's repository per saved keystroke, against a daily limit.
- **Why `prev_base_sha` must never be copied onto the first rule.** A `{{ ... }}` template over a field the event does not have renders to the literal text `<no value>`, which is rejected — so that one line would stop **every** check in the organization, not just retargets. This is the same class of trap as the `pr`-is-a-path warning already on the page, and a reader who has just learned "every other field is a template" is exactly the reader who would hit it.

Also updated: the "draft pull requests are checked" note and the "everything else is ignored" paragraph now describe the action set as it actually is.

`markdownlint-cli2` over `docs/**/*.md`: `Summary: 0 issues in 0 files`.

## Verified, not assumed

The behaviour was reproduced end to end against a test organization before this was written: a pull request whose diff introduced nothing got a green check; retargeting it at a base that did not carry its dependency left that green check standing while the pull request introduced 11 HIGH findings; and a title-only edit produced no activity anywhere. The payload shapes quoted (`changes.base.sha.from` for a retarget, `changes.title.from` for a title edit, and the `<no value>` rendering) were each read off a real delivery or run, not taken from documentation.


lcbill (dismissed):

lcbill (dismissed):

maximelb:
**Self-review, and a deliberate HOLD.** The `/lc-review` bot has not reviewed this repository's PRs (requested 05:14Z and 13:54Z, nothing posted), so per the standing rule I ran an independent reviewer scoped to this worktree.

Addressed in `2db4d98`: the warning's reasoning was wrong (a rule *can* nest an `and` in its `or`, and an `edited` request with no previous base is refused before any scan — the real reason is that a rule has one response); the "Is it firing?" replay now covers the retarget rule; the refusal list gains "the new base is the same commit"; "replaces the conclusion" became "a newer check run supersedes". YAML untouched, so it stays identical to the recipe module and the internal copy. markdownlint: 0 issues.

**Not merging yet, on the reviewer's HIGH finding, which is right:** this is the public page, and the rule it tells customers to install cannot work until the extension that declares `prev_base_sha` and the API that accepts `edited` are both released and deployed. Merged now, it would remove a note that is still true and send readers to a rule that is refused on every retarget. It should merge together with that rollout.

lcbill (approved):

PR #418, opened 2026-09-16
Email Security: explain the history a provider will not give us

maximelb:
The historical backfill cannot retrieve everything that was already in a mailbox — mail deleted between the provider listing it and us asking for it, and mail above the provider API's own size ceiling. That is normal, and until now nothing told a customer it had happened or how much.

It is counted on the coverage screen rather than reported as an ingest error. `EMAIL_INGEST_ERROR` means *"this message will never be ingested"* and exists so a mailbox quietly falling out of coverage is visible; a first connection over a large estate produces unretrievable history in the thousands, and putting those on the same feed would bury every real coverage loss under mail nobody could have read.

Adds a **Some of the history never arrives** section to Troubleshooting, beside the existing `EMAIL_INGEST_ERROR` section so the two are read against each other, plus a Quick-reference row. It documents:

- the two new `coverage.backfill` fields, `messages_skipped` and `mailboxes_skipped`;
- why they are read as a **pair** — the same total is an ordinary size ceiling when spread thin and one unreadable mailbox when concentrated, and those need different responses;
- that they describe the **current walk**: if a mailbox's history walk restarts, the count restarts with it rather than accumulating across attempts;
- that a mailbox which has not finished a page of history is in neither number — unmeasured is not the same as having lost nothing;
- that skipped history does **not** hold `backfill.complete` open, since a page containing a message the provider refuses has still been walked.

**Self-reviewed rather than `/lc-review`'d**, per the convention for this repo. What that review changed: the first draft said an unmeasured mailbox "reports nothing at all rather than zero", which pointed at a per-mailbox field that is not in the customer-visible payload — the coverage response is organization-level. Corrected to say what such a mailbox does to the two totals, which is what a reader can actually act on. Also checked: no vendor is named, no internal identifiers appear, and every claim matches the behaviour being shipped rather than the behaviour proposed.

**Do not merge yet** — this documents fields that are not in production. It merges with the production release.


maximelb:
merge with the prod release

lcbill (dismissed):

lcbill (dismissed):

lcbill (approved):

PR #419, opened 2026-09-16
docs(harmony): stop the Authentication URL from becoming a doubled path

pllesperance-lc:
> [!IMPORTANT]
> **Do not merge before refractionPOINT/usp-adapters#320.** This page documents the adapter behaviour that PR implements — the startup `url` rejection, the 403 soft-fail, and the all-services-refused error. Merged first, these statements are false for every adapter currently deployed.

## What was asked

The Check Point Harmony adapter page walks the reader through creating an Infinity Portal API key. Following step 5 literally produces a broken adapter. Fix the wording, and document how the gateway reports a cloud service the tenant cannot read.

## What was done

**The `url` field.** Step 5 said to take the **Authentication URL** shown next to a new API key and supply it as the adapter's `url`. That URL is displayed in full, ending in `/auth/external`, and the adapter appends `/auth/external` itself, so following the instruction gives `Cannot POST /auth/external/auth/external` with nothing pointing back at the `url` field. Step 5 now carries a warning admonition with the correct and incorrect forms, and says authentication is the only place the mistake ever surfaces — it runs first, so the data calls never get far enough to fail on their own mangled paths.

The `url` option description states the rule exactly: a path containing one of the adapter's own API segments (`/auth/external`, `/app/laas-logs-api`, `/app/hec-api`, matched case-insensitively and on segment boundaries), or a query or fragment, is rejected at startup — while an unrelated reverse-proxy prefix, including one that merely resembles ours such as `/auth/external-gw`, stays valid.

**Unlicensed and unknown cloud services.** The page documented only the `Canceled` state. It now has a `#### When a cloud service can't be read` section covering both that and the 403 refusal, what a soft failure costs (the skipped window is not retried), the two causes worth telling apart, and — in an admonition — every case that is reported as an error instead: a 403 arriving after the submit, a 403 that is not the gateway's own refusal, all configured services being refused at once, and a misspelled service name.

`docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md`.

## Testing

Markdown-only change. CI is green, including **Build rendered docs**, which exercises the actual MkDocs render — the `!!! warning` nested inside ordered-list item 5 and the `!!! note` with its bullets both render correctly.

## Review

Four adversarial review rounds. Findings addressed: the tense contradiction between the admonition and the option description; a synthesized `HTTP 403: …` string that is emitted nowhere; an overbroad "any path is rejected" claim that would have told reverse-proxy operators their working config was invalid; "neither surfaces as an error" being wrong for a 403 after the submit; the soft-fail rule being the gateway's exact phrase rather than "names the cloud service", which put the misspelled-service-name case on the page; the never-retried window; the wrong 404 mechanism; and the section's size and placement.


pllesperance-lc (commented):
Findings (inline threads):

- L230: the admonition says pasting the full Authentication URL "makes every request double the suffix", which contradicts L31 ("A `url` with a path is rejected at startup") — with the startup check in place that 404 is unreachable.
- L47: `HTTP 403: Unauthorized to perform operations on the given Cloud Service` is quoted as a literal, but the adapter does not emit that exact string, so grepping for it finds nothing.
- L47/L49: the two causes are not equivalent — an API key missing *Logs as a Service* fails **every** cloud service, so "trim `cloud_services`" silences a silently dead Events source instead of fixing it.
- L49: "Neither surfaces as an error" holds only for a 403 on the query submit; a 403 on the status poll or record retrieval is still a hard error.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
```diff
@@ -219,7 +224,17 @@ harmony:
     - *Harmony Email & Collaboration* for the Entities source.
     - A single key with both services attached is fine.
 4. Copy the resulting **Client ID** and **Access Key**. The Access Key is shown only once — save it somewhere safe.
-5. Note the **Authentication URL** shown next to the key. If it points at a regional gateway (`cloudinfra-gw-us.portal.checkpoint.com`, `cloudinfra-gw-eu.portal.checkpoint.com`, etc.) you will need to supply that hostname as the adapter's `url` value.
+5. Note the **Authentication URL** shown next to the key. If it points at a regional gateway (`cloudinfra-gw-us.portal.checkpoint.com`, `cloudinfra-gw-eu.portal.checkpoint.com`, etc.) you will need to supply that gateway as the adapter's `url` value.
+
+    !!! warning "Use only the scheme and host — drop the `/auth/external` suffix"
+        The Authentication URL is shown in full, ending in `/auth/external`, but the adapter's `url` field takes only the part before that. The adapter appends `/auth/external` (and the other API paths) itself, so pasting the whole value makes every request double the suffix:
```
pllesperance-lc: This contradicts line 31, which says a `url` with a path "is rejected at startup". Both are in the present tense, but for a given adapter build only one can be true: if `Validate` refuses the path, the reader never reaches a request, so `Cannot POST /auth/external/auth/external` is unreachable. And a reader who *does* hit that 404 will conclude the startup check did not fire and go hunting somewhere else.

Pick one story — describe what the reader actually gets now, and keep the doubled path only as the reason the check exists:

> The Authentication URL is shown in full, ending in `/auth/external`, but the adapter's `url` field takes only the part before that — the adapter appends `/auth/external` (and the other API paths) itself. Pasting the whole value would make every request double the suffix (`POST /auth/external/auth/external` → 404), so the adapter refuses to start and tells you the value to use instead.

If the startup rejection is new adapter behaviour rather than long-standing, that is worth a word too: an adapter already running an older build still fails the old way, and this page is the first thing its operator will read.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
```diff
@@ -41,7 +41,12 @@ All duration fields below are parsed with [`time.ParseDuration`](https://pkg.go.
 - `events.page_limit` *(optional)*: page size for the records-retrieval API. Defaults to `100`. The gateway rejects values below `10` with HTTP 400.
 - `events.limit` *(optional)*: cap on records returned per cloud service per poll. Defaults to `5000`.
 
-If a configured `cloud_service` is not provisioned for the tenant the gateway returns the query in state `Canceled`; the adapter logs one warning per poll and keeps going (it does not surface as an error). Remove the service from `cloud_services` to silence the warning.
+`events.cloud_services` defaults to the **full** Harmony suite, so a tenant licensed for only part of it will be queried for products it cannot read. The gateway reports that in one of two ways, and the adapter treats both as a per-service soft failure — one warning per poll, that window skipped, the other cloud services unaffected:
+
+- the query comes back in state `Canceled`; or
+- the query is rejected with `HTTP 403: Unauthorized to perform operations on the given Cloud Service`, meaning the tenant is not licensed for that product, or the API key is missing the *Logs as a Service* service.
```
pllesperance-lc: Two issues with this bullet.

**1. The backticked string is not what the adapter emits.** The gateway's 403 body is JSON and the sentence arrives inside `error.details[]`; the warning the adapter raises names the offending cloud service and carries that body, and it is not prefixed `HTTP 403: `. Presenting a synthesised composite in backticks invites an operator to grep their adapter warnings for it and find nothing. Describe it instead of quoting it, e.g. "…is rejected with HTTP 403, carrying `Unauthorized to perform operations on the given Cloud Service`".

**2. The two causes are not equivalent, and the bullet treats them as if they were.** "Tenant is not licensed for that product" is genuinely per-service, and narrowing `cloud_services` (L49) is the right fix. "API key is missing the *Logs as a Service* service" is **not** per-service — it fails every configured cloud service, so the whole Infinity Events source ships nothing while reporting no error at all, and the L49 remedy silences the only symptom without fixing anything. Worth stating plainly: if *every* configured service warns, the API key is the problem; fix the key rather than trimming the list.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
```diff
@@ -41,7 +41,12 @@ All duration fields below are parsed with [`time.ParseDuration`](https://pkg.go.
...
+
+Neither surfaces as an error. Set `events.cloud_services` to just the products the tenant is licensed for to silence the warnings.
```
pllesperance-lc: "Neither surfaces as an error" is true only of a 403 on the **query submit**. A 403 returned later in the sequence — the task status poll or the record retrieval — is still a hard error, and has to be: records from that window may already be downstream, so the window cannot simply be skipped. As written this page tells an operator that any Harmony 403 is benign, which is exactly the wrong conclusion in the one case where data is being lost.

Suggest qualifying, e.g.:

> Neither surfaces as an error when it happens on the query submit. A 403 on the subsequent status poll or record retrieval is still reported as an error — records from that window may already have shipped, so it is not safe to skip.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
pllesperance-lc: Fixed in 8909b954: the admonition now leads with the startup refusal and names the doubled-path 404 as what older builds do instead, so it no longer contradicts the option description.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
pllesperance-lc: Fixed in 8909b954: the bullet now says the sentence arrives in `error.details` in the 403 body rather than quoting a composite string that is emitted nowhere. The two causes are also split out — tenant not licensed refuses one service, a key missing the Logs as a Service grant refuses all of them and silently ships nothing, which is now called out as its own symptom with the key as the fix.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
pllesperance-lc: Fixed in 8909b954: dropped the blanket claim and added a note that only a 403 on the query submission is soft-failed; a 403 during status polling or record retrieval stays an error because records from that window may already be downstream.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
```diff
@@ -28,7 +28,7 @@ Adapter Type: `harmony`
...
 - `access_key`: Infinity Portal Access Key paired with the Client ID above.
-- `url` *(optional)*: Infinity Portal gateway base URL. Defaults to `https://cloudinfra-gw.portal.checkpoint.com`. Use the regional variant (for example `https://cloudinfra-gw-us.portal.checkpoint.com`) if your tenant lives in a regional data center. Both `/app/laas-logs-api` and `/app/hec-api` share the same hostname per region.
+- `url` *(optional)*: Infinity Portal gateway base URL — **scheme and host only**, with no path. Defaults to `https://cloudinfra-gw.portal.checkpoint.com`. Use the regional variant (for example `https://cloudinfra-gw-us.portal.checkpoint.com`) if your tenant lives in a regional data center. Both `/app/laas-logs-api` and `/app/hec-api` share the same hostname per region. Do not paste the Infinity Portal's **Authentication URL** here: it ends in `/auth/external`, which the adapter appends itself. A `url` with a path is rejected at startup.
```
pllesperance-lc: "**scheme and host only**, with no path" and "A `url` with a path is rejected at startup" are both broader than what the adapter actually enforces, and the overreach costs a real deployment shape.

`validateGatewayURL` refuses a `url` only when its path **contains one of the adapter's own API segments** (`/auth/external`, `/app/laas-logs-api`, `/app/hec-api`), or when it carries a query or fragment. An unrelated path prefix is *deliberately* accepted — that is the gateway reached through a reverse proxy, e.g. `https://proxy.example.com/checkpoint`, which composes correctly and has a test pinning it so upgrades don't take working deployments down. As written, this line tells an operator running exactly that setup that their config is invalid and will be refused at startup.

Suggest describing the actual rule (it also documents the query/fragment case, which is currently undocumented):

> - `url` *(optional)*: Infinity Portal gateway base URL. Defaults to `https://cloudinfra-gw.portal.checkpoint.com`. Use the regional variant (for example `https://cloudinfra-gw-us.portal.checkpoint.com`) if your tenant lives in a regional data center. Both `/app/laas-logs-api` and `/app/hec-api` share the same hostname per region. Do not paste the Infinity Portal's **Authentication URL** here: it ends in `/auth/external`, which the adapter appends itself. A `url` whose path already contains one of the paths the adapter appends (`/auth/external`, `/app/laas-logs-api`, `/app/hec-api`), or that carries a query string or fragment, is rejected at startup with an error naming the value to use instead. An unrelated path prefix is fine — use it when the gateway is reached through a reverse proxy (`https://proxy.example.com/checkpoint`).

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
```diff
@@ -219,7 +232,15 @@ harmony:
...
+5. Note the **Authentication URL** shown next to the key. If it points at a regional gateway (`cloudinfra-gw-us.portal.checkpoint.com`, `cloudinfra-gw-eu.portal.checkpoint.com`, etc.) you will need to supply that gateway as the adapter's `url` value.
+
+    !!! warning "Use only the scheme and host — drop the `/auth/external` suffix"
+        The Authentication URL is shown in full, ending in `/auth/external`, but the adapter's `url` field takes only the part before that — the adapter appends `/auth/external` and the other API paths itself.
+
+        - Correct: `https://cloudinfra-gw-eu.portal.checkpoint.com`
+        - Incorrect: `https://cloudinfra-gw-eu.portal.checkpoint.com/auth/external`
+
+        A `url` carrying a path is refused when the adapter starts, with an error naming the value to use instead. On adapter builds predating that check the value is accepted and every request doubles the suffix, which fails as `Cannot POST /auth/external/auth/external`.
```
pllesperance-lc: Same overreach as the `url` option description (separate thread on L31): "A `url` carrying a path is refused when the adapter starts" is not what ships. Only a path repeating one of the adapter's own API segments is refused; a reverse-proxy prefix such as `https://proxy.example.com/checkpoint` stays valid by design. An operator behind a proxy reads this step, matches "carrying a path", and concludes their working gateway URL is about to be rejected.

The admonition title has the same problem — "Use only the scheme and host" is advice the validator does not back.

Suggest:

> !!! warning "Drop the `/auth/external` suffix from the Authentication URL"
>     The Authentication URL is shown in full, ending in `/auth/external`, but the adapter's `url` field takes only the part before that — the adapter appends `/auth/external` and the other API paths itself.
>
>     - Correct: `https://cloudinfra-gw-eu.portal.checkpoint.com`
>     - Incorrect: `https://cloudinfra-gw-eu.portal.checkpoint.com/auth/external`
>
>     A `url` that already contains one of the paths the adapter appends is refused when the adapter starts, with an error naming the value to use instead. (A path prefix that isn't one of those — a gateway behind a reverse proxy — is accepted.) On adapter builds predating that check the value is accepted and the auth call doubles the suffix, failing as `Cannot POST /auth/external/auth/external`.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
```diff
@@ -41,7 +41,20 @@ All duration fields below are parsed with [`time.ParseDuration`](https://pkg.go.
...
+
+Set `events.cloud_services` to just the products the tenant is licensed for to silence the warnings.
+
+A 403 has two quite different causes, and they are worth telling apart before narrowing the list:
+
+- **The tenant is not licensed for that product.** Only that cloud service is refused; the rest keep ingesting normally.
+- **The API key is missing the *Logs as a Service* service.** *Every* cloud service is refused, so the Infinity Events source ships nothing at all. If all of the configured services warn at once, suspect the key rather than the licensing, and check the key's attached services in the Infinity Portal.
+
+!!! note "Only a refused query submission is soft-failed"
+    A 403 on the *submission* of a query is treated as a warning, as described above. A 403 that arrives later, while the adapter is polling the query's status or retrieving its records, is reported as an error instead — records from that window may already have been shipped, so the adapter must not quietly skip past the remainder.
```
pllesperance-lc: The submit-vs-later split is only half the condition, and the half that's missing is the one an operator is most likely to hit.

A submit 403 is soft-failed **only when the response body names the cloud service** (`isCloudServiceForbidden`). A 403 on the submit that doesn't — an IP allowlist on the key, or a corporate proxy/WAF in front of the gateway returning an HTML 403 — stays a hard error, deliberately and with a test pinning it. This page already points operators at reverse-proxied gateways elsewhere, so "a 403 on the submission is treated as a warning" will be read by exactly the audience for whom it is false.

The title has the converse problem: `Canceled` is soft-failed too and it is not a refused submission, it's a state on the status poll.

Suggest:

> !!! note "Which 403s are soft-failed"
>     A 403 is downgraded to a warning only when it refuses the *submission* of a query **and** its body names the cloud service, as above. Any other 403 on the submit — an IP restriction on the API key, or a proxy or WAF in front of the gateway — is reported as an error, so a blocked deployment stays loud instead of looking like an unlicensed product. A 403 that arrives later, while the adapter is polling the query's status or retrieving its records, is likewise an error: records from that window may already have been shipped, so the adapter must not quietly skip past the remainder.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
```diff
@@ -41,7 +41,20 @@ All duration fields below are parsed with [`time.ParseDuration`](https://pkg.go.
...
+
+- the query comes back in state `Canceled`; or
+- the query is refused with HTTP 403, whose body carries `Unauthorized to perform operations on the given Cloud Service` in `error.details`.
+
+Set `events.cloud_services` to just the products the tenant is licensed for to silence the warnings.
+
+A 403 has two quite different causes, and they are worth telling apart before narrowing the list:
+
+- **The tenant is not licensed for that product.** Only that cloud service is refused; the rest keep ingesting normally.
+- **The API key is missing the *Logs as a Service* service.** *Every* cloud service is refused, so the Infinity Events source ships nothing at all. If all of the configured services warn at once, suspect the key rather than the licensing, and check the key's attached services in the Infinity Portal.
```
pllesperance-lc: The all-services-refused case is described as something the operator has to infer ("If all of the configured services warn at once, suspect the key"), but the adapter no longer leaves it to them: once *every* configured cloud service has been refused with a cloud-service 403, it raises one error of its own — that the source is ingesting nothing and that the API key is most likely missing the *Logs as a Service* grant. It is latched, so it fires once per episode rather than every poll, and re-arms as soon as any service becomes reachable again.

Two consequences for the text:

- L44's "the adapter treats both as a per-service soft failure" needs an exception, otherwise this page says no error is ever raised for a 403 while the shipped adapter raises one. Suggest appending: "…the other cloud services unaffected. The one exception is *every* configured service being refused with a 403 — see below."
- This bullet should say the adapter reports it, not that the operator should spot the pattern. Also worth keeping the second cause the adapter's own message hedges toward, since a tenant licensed for none of the default suite lands here too:

> - **The API key is missing the *Logs as a Service* service.** *Every* cloud service is refused, so the Infinity Events source ships nothing at all. The adapter raises a single error when it sees every configured service refused, naming this as the likely cause — fix the key's attached services in the Infinity Portal rather than trimming `cloud_services`. (The same error appears if the key is fine but the tenant is licensed for none of the configured products; in that case narrowing the list is the fix.)

pllesperance-lc:
Second-round review. The three earlier findings are genuinely addressed; the remaining ones are new, and all four come from the companion adapter change having moved since round one.

- L31 / L243: "no path" / "a `url` carrying a path is refused" is broader than `validateGatewayURL` — only a path repeating the adapter's own API segments (`/auth/external`, `/app/laas-logs-api`, `/app/hec-api`) is refused; a reverse-proxy prefix is accepted on purpose. Query/fragment rejection is also undocumented.
- L56-57: a submit 403 is soft-failed only when the body names the cloud service — a proxy/WAF/IP-restriction 403 stays a hard error, which is the case this wording turns benign. Title also misses `Canceled`, which is soft-failed and isn't a submit refusal.
- L44 / L54: every-service-refused is now a single latched error from the adapter, not a pattern the operator has to spot in the warnings; L44's blanket "soft failure" needs that exception.

No rendering issues (4/8-space admonition nesting and backticked titles match repo precedent) and nothing internal leaked.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
pllesperance-lc: Fixed in 3d8bf201. The bullet now states the actual rule — a url already containing `/auth/external`, `/app/laas-logs-api` or `/app/hec-api`, or carrying a query or fragment, is rejected — and says explicitly that an unrelated prefix is fine, with the reverse-proxy example. Good catch: the docs were written against the first version of the check and the code narrowed underneath them.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
pllesperance-lc: Fixed in 3d8bf201: title is now "Drop the `/auth/external` suffix" and the body says "a `url` that already contains one of those API paths" rather than any path.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
pllesperance-lc: Fixed in 3d8bf201. The note is retitled "When a 403 is *not* treated as a warning" and gives both exclusions as separate bullets — the later-in-the-sequence one, and the body-must-name-the-cloud-service one with the IP restriction / proxy / WAF cases spelled out. Retitling also resolves the `Canceled` overlap, since the note no longer claims to characterise everything that is soft-failed.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
pllesperance-lc: Fixed in 3d8bf201: the bullet now says the adapter raises a single error for the all-refused case instead of leaving it to be inferred, and the intro sentence is scoped to "each is treated as a per-service soft failure" with that exception stated where it belongs. Also took your point on the hedge — the bullet now carries the second cause (the configured products being ones the tenant has none of), matching what the adapter's own message says.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
```diff
@@ -28,7 +28,7 @@ Adapter Type: `harmony`
...
-- `url` *(optional)*: Infinity Portal gateway base URL. Defaults to `https://cloudinfra-gw.portal.checkpoint.com`. Use the regional variant (for example `https://cloudinfra-gw-us.portal.checkpoint.com`) if your tenant lives in a regional data center. Both `/app/laas-logs-api` and `/app/hec-api` share the same hostname per region.
+- `url` *(optional)*: Infinity Portal gateway **base** URL — the API paths are appended to it. Defaults to `https://cloudinfra-gw.portal.checkpoint.com`. Use the regional variant (for example `https://cloudinfra-gw-us.portal.checkpoint.com`) if your tenant lives in a regional data center. Both `/app/laas-logs-api` and `/app/hec-api` share the same hostname per region. Do not paste the Infinity Portal's **Authentication URL** here: it ends in `/auth/external`, which the adapter appends itself. A `url` that already contains one of the adapter's own API paths (`/auth/external`, `/app/laas-logs-api`, `/app/hec-api`), or that carries a query or fragment, is rejected at startup. An unrelated path prefix is fine — if you reach the gateway through a reverse proxy, `https://proxy.example.com/checkpoint` is a valid value.
```
pllesperance-lc: "already contains one of the adapter's own API paths" is broader than `validateGatewayURL`. The match is on whole path segments and case-insensitive, so a reverse-proxy prefix that merely starts with one of them — `/auth/external-gw`, `/app/hec-api-mirror` — is deliberately accepted. As written this sentence says such a URL is rejected while the next sentence says an unrelated prefix is fine, leaving the two rules in conflict for exactly the prefixes that look like ours. Suggest: "A `url` whose path contains one of the adapter's own API paths as a whole segment (`/auth/external`, `/app/laas-logs-api`, `/app/hec-api`, matched case-insensitively), or that carries a query or fragment, is rejected at startup.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
```diff
@@ -41,7 +41,23 @@ All duration fields below are parsed with [`time.ParseDuration`](https://pkg.go.
 - `events.page_limit` *(optional)*: page size for the records-retrieval API. Defaults to `100`. The gateway rejects values below `10` with HTTP 400.
 - `events.limit` *(optional)*: cap on records returned per cloud service per poll. Defaults to `5000`.
 
-If a configured `cloud_service` is not provisioned for the tenant the gateway returns the query in state `Canceled`; the adapter logs one warning per poll and keeps going (it does not surface as an error). Remove the service from `cloud_services` to silence the warning.
+`events.cloud_services` defaults to the **full** Harmony suite, so a tenant licensed for only part of it will be queried for products it cannot read. The gateway reports that in one of two ways, and each is treated as a per-service soft failure — one warning per poll, that window skipped, the other cloud services unaffected:
```
pllesperance-lc: "that window skipped" understates the consequence. On both soft failures the worker sets `nextStart = endTime`, so the skipped window is never re-queried: a refusal that turns out to be transient permanently loses that service's events for that window, with only a warning to show for it. The rest of the section frames these warnings as benign noise to silence, so the trade-off should be explicit — e.g. "… one warning per poll, that window skipped (the adapter does not come back for it, so a refusal that turns out to be transient costs that window's events), the other cloud services unaffected".

pllesperance-lc (commented):
Third round. Four open findings, all page wording against the current adapter behaviour:

- L57/L60: the soft-fail needs the gateway's exact phrase, not a body that "names the cloud service" — a misspelled service name (`The provided Cloud Service is unknown`) stays a hard error, and the listed causes for a hard 403 omit it.
- L51: "A 403 has two quite different causes" contradicts the admonition at L56-60, which lists more.
- L44: "that window skipped" should say the window is never re-queried — a transient refusal silently costs that window's events.
- L31: "already contains one of the adapter's own API paths" is broader than the segment-bounded check, which accepts `/auth/external-gw` and `/app/hec-api-mirror` prefixes.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
```diff
@@ -41,7 +41,23 @@ All duration fields below are parsed with [`time.ParseDuration`](https://pkg.go.
...
+
+- the query comes back in state `Canceled`; or
+- the query submission is refused with HTTP 403, whose body carries `Unauthorized to perform operations on the given Cloud Service` in `error.details`.
+
+Set `events.cloud_services` to just the products the tenant is licensed for to silence the warnings.
+
+A 403 has two quite different causes, and they are worth telling apart before narrowing the list:
```
pllesperance-lc: "A 403 has two quite different causes" is contradicted by the admonition four lines down, which lists an IP restriction and a proxy/WAF as further causes (and a misspelled service name is another). Scope it to the soft-failed case: "A *soft-failed* 403 has two quite different causes, and they are worth telling apart before narrowing the list:".

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
```diff
@@ -41,7 +41,23 @@ All duration fields below are parsed with [`time.ParseDuration`](https://pkg.go.
...
+
+- **The tenant is not licensed for that product.** Only that cloud service is refused; the rest keep ingesting normally.
+- **The API key is missing the *Logs as a Service* service.** *Every* cloud service is refused, so the Infinity Events source ships nothing at all. Rather than let that pass as a handful of warnings, the adapter raises a single error saying every configured cloud service was refused. Check the key's attached services in the Infinity Portal — and if the key is right, the configured products are ones this tenant has none of.
+
+!!! note "When a 403 is *not* treated as a warning"
+    Only a 403 refusing the **submission** of a query, and only one whose body names the cloud service, is soft-failed.
```
pllesperance-lc: `whose body names the cloud service` is not the rule the adapter implements. `isCloudServiceForbidden` soft-fails only a body containing the exact sentence `Unauthorized to perform operations on the given Cloud Service` (case-insensitive substring); anything else is a hard error. Two cases this wording gets backwards:

- `The provided Cloud Service is unknown` — the gateway's answer to a *misspelled* service name, i.e. the `and` vs `&` mistake warned about on L38 — does name the cloud service and is still a hard error, erroring every poll with that service never ingesting.
- If Check Point rewords the message, every refusal falls back to a hard error.

So L60's diagnosis ("It is not the gateway declining one product but something else refusing the request: an IP restriction, or a corporate proxy or WAF in front of the gateway") is wrong for both, and the misspelled-name case — the most likely one for an operator hand-editing `events.cloud_services` — is nowhere on the page. Suggested replacement for L57/L60:

> Only a 403 refusing the **submission** of a query, and only one whose body carries the gateway's exact phrase `Unauthorized to perform operations on the given Cloud Service`, is soft-failed.
>
> - *(status-poll / retrieve bullet unchanged)*
> - Any other 403 is reported as an error. That covers an IP restriction or a corporate proxy or WAF in front of the gateway — but also a **misspelled cloud service name**, which the gateway refuses with `The provided Cloud Service is unknown`. That one is not a licensing gap and will not go away on its own: it errors every poll until the spelling in `events.cloud_services` is fixed (see the ampersand note above).

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
pllesperance-lc: Fixed in a9416a55, using your wording — "as a whole segment (`/auth/external`, `/app/laas-logs-api`, `/app/hec-api`, matched case-insensitively)" — and the following sentence now names `/auth/external-gw` explicitly as an accepted example, so the two rules can't read as being in conflict.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
pllesperance-lc: Fixed in a9416a55: added a sentence saying the skipped window is not retried and that a transient refusal therefore costs that cloud service those events, framed as the deliberate trade against pinning the cursor. Agreed the section read as though the warnings were pure noise.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
pllesperance-lc: Fixed in a9416a55: now "A *soft-failed* 403 has two quite different causes".

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
pllesperance-lc: Fixed in a9416a55. The note now gives the exact phrase and says any other 403 is an error, and the misspelled-name case is called out as its own paragraph tied back to the ampersand warning — including that it errors every poll until the spelling is fixed.

One deviation from your suggested text: I kept the misspelling out of the "any other 403" bullet and put it in its own paragraph, because the source doesn't establish that `The provided Cloud Service is unknown` comes back as a 403 specifically — the adapter only records the message. The operator-facing point (hard error, never self-resolves, fix the spelling) is the same either way, without asserting a status code I can't back up.

pllesperance-lc (commented):
Fourth round. Four open findings:

- L31 (and L250): the page states behaviour no released adapter has — the startup `url` check and the 403 soft-fail are still an open **draft** in refractionPOINT/usp-adapters#320.
- L250: "every request doubles the suffix" is wrong — only the token call doubles `/auth/external`; data calls get a different, non-doubled bad path and never run.
- L44: the 403 material is ~20 unheaded lines of failure-mode prose inside the config-field reference, duplicating L40 and repeating the "narrow the list" remedy three times.
- L58: the "when a 403 is *not* a warning" admonition omits the all-services-refused error from L56, which also contradicts the blanket "soft failure" framing at L44/L53.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
```diff
@@ -28,7 +28,7 @@ Adapter Type: `harmony`
...
+- `url` *(optional)*: Infinity Portal gateway **base** URL — the API paths are appended to it. Defaults to `https://cloudinfra-gw.portal.checkpoint.com`. Use the regional variant (for example `https://cloudinfra-gw-us.portal.checkpoint.com`) if your tenant lives in a regional data center. Both `/app/laas-logs-api` and `/app/hec-api` share the same hostname per region. Do not paste the Infinity Portal's **Authentication URL** here: it ends in `/auth/external`, which the adapter appends itself. A `url` whose path contains one of the adapter's own API paths as a whole segment (`/auth/external`, `/app/laas-logs-api`, `/app/hec-api`, matched case-insensitively), or that carries a query or fragment, is rejected at startup. An unrelated path prefix is fine — if you reach the gateway through a reverse proxy, `https://proxy.example.com/checkpoint` is a valid value, as is a prefix that merely resembles one of ours such as `/auth/external-gw`.
```
pllesperance-lc, line 31: The startup rejection described here does not exist in any released adapter yet: `validateGatewayURL`, and the 403 soft-fail documented at L44-L64, both live in refractionPOINT/usp-adapters#320, which is still an open draft. Merging this page first tells an operator their running adapter refuses a bad `url` and warn-and-skips an unlicensed cloud service, when today it does neither — a `url` with a path is accepted and 404s, and a 403 on submit is a hard error every poll with the cursor pinned.

Fix: hold this PR until #320 ships, or scope the new claims ("is rejected at startup", "treated as a per-service soft failure", "the adapter raises a single error") to the adapter version that introduces them.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
```diff
@@ -219,7 +239,15 @@ harmony:
...
+5. Note the **Authentication URL** shown next to the key. If it points at a regional gateway (`cloudinfra-gw-us.portal.checkpoint.com`, `cloudinfra-gw-eu.portal.checkpoint.com`, etc.) you will need to supply that gateway as the adapter's `url` value.
+
+    !!! warning "Drop the `/auth/external` suffix"
+        The Authentication URL is shown in full, ending in `/auth/external`, but the adapter's `url` field takes only the part before that — the adapter appends `/auth/external` and the other API paths itself.
+
+        - Correct: `https://cloudinfra-gw-eu.portal.checkpoint.com`
+        - Incorrect: `https://cloudinfra-gw-eu.portal.checkpoint.com/auth/external`
+
+        A `url` that already contains one of those API paths is refused when the adapter starts, with an error naming the value to use instead. On adapter builds predating that check it is accepted and every request doubles the suffix, which fails as `Cannot POST /auth/external/auth/external`.
```
pllesperance-lc: "every request doubles the suffix" is not what happens. With `url` = `https://host/auth/external`, only the token call doubles the segment (`conf.URL + authPath` → `/auth/external/auth/external`); a data call is `conf.URL + eventsQueryPath` → `https://host/auth/external/app/laas-logs-api/api/logs_query`, which is broken but not doubled. And since the token call is issued first, it is the only failure an operator ever sees.

Secondly, "builds predating that check" is, today, every build (see the L31 thread), so the sentence reads as if the check were already the norm.

Fix: "The value is accepted, and the first call the adapter makes — the token request — goes to `/auth/external/auth/external`, which fails with `Cannot POST /auth/external/auth/external` before any events are fetched."

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
```diff
@@ -41,7 +41,27 @@ All duration fields below are parsed with [`time.ParseDuration`](https://pkg.go.
 - `events.page_limit` *(optional)*: page size for the records-retrieval API. Defaults to `100`. The gateway rejects values below `10` with HTTP 400.
 - `events.limit` *(optional)*: cap on records returned per cloud service per poll. Defaults to `5000`.
 
-If a configured `cloud_service` is not provisioned for the tenant the gateway returns the query in state `Canceled`; the adapter logs one warning per poll and keeps going (it does not surface as an error). Remove the service from `cloud_services` to silence the warning.
+`events.cloud_services` defaults to the **full** Harmony suite, so a tenant licensed for only part of it will be queried for products it cannot read. The gateway reports that in one of two ways, and each is treated as a per-service soft failure — one warning per poll, that window skipped, the other cloud services unaffected:
```
pllesperance-lc: This block has grown to ~20 lines of failure-mode prose sitting unheaded between the `events` bullets and the `entities` block, inside what is otherwise a config-field reference. It buries the `entities` options for someone reading the reference, and it is not where someone debugging a live adapter would look.

It also restates itself: L40 already says `cloud_services` defaults to the full suite, so this lead-in repeats it; and the "narrow the list" remedy is given three times (here, again as the standalone L51 sentence, and again inside L56).

Fix: give it a heading in the page's existing style — `#### Unlicensed or unknown cloud services`, alongside `#### Two cursor modes` / `#### Annotations` — drop the duplicated "defaults to the full suite" lead-in and the standalone L51 sentence.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
```diff
@@ -41,7 +41,27 @@ All duration fields below are parsed with [`time.ParseDuration`](https://pkg.go.
...
+A skipped window is not retried: the adapter advances past it and moves on, so a refusal that turns out to be transient costs that cloud service the events in that window. That is the deliberate trade for not pinning the cursor and re-querying an ever-growing window on a product the tenant will never be able to read.
+
+Set `events.cloud_services` to just the products the tenant is licensed for to silence the warnings.
+
+A *soft-failed* 403 has two quite different causes, and they are worth telling apart before narrowing the list:
+
+- **The tenant is not licensed for that product.** Only that cloud service is refused; the rest keep ingesting normally.
+- **The API key is missing the *Logs as a Service* service.** *Every* cloud service is refused, so the Infinity Events source ships nothing at all. Rather than let that pass as a handful of warnings, the adapter raises a single error saying every configured cloud service was refused. Check the key's attached services in the Infinity Portal — and if the key is right, the configured products are ones this tenant has none of.
+
+!!! note "When a 403 is *not* treated as a warning"
```
pllesperance-lc: This admonition promises the cases where a 403 is *not* just a warning, but omits the one an operator is most likely to be staring at: every configured cloud service refused, which the adapter escalates to a single error — described four lines earlier at L56, in a bullet list introduced as "a *soft-failed* 403" (L53) under a paragraph that says both forms are "treated as a per-service soft failure — one warning per poll" (L44). So the escalation is both mislabelled as a soft failure and missing from the list that exists to enumerate it.

Fix: qualify L44 ("…unless every configured cloud service is refused, which is escalated to a single error") and move L56's escalation into this admonition, so every not-just-a-warning case is in one place.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
pllesperance-lc: Real risk, and thanks for checking the PR state rather than assuming. Handled by making the dependency a blocking note at the top of the PR description: this must not merge before refractionPOINT/usp-adapters#320.

I deliberately did not hedge the prose itself. Version caveats on the page would have to be found and removed once the adapter ships, and they tend not to be — the page would be permanently vaguer for a transient reason. Documenting a feature in the present tense and landing it with the feature is the normal arrangement; the merge order is a reviewer decision, and it is now impossible to miss.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
pllesperance-lc: Fixed in ff7ae58f. You're right — only `conf.URL + authPath` repeats the segment; a data call gets a differently mangled path, and since `getToken` runs first it is the only failure anyone ever sees. The admonition now says exactly that, which is also the more useful thing to tell someone staring at the 404. Dropped the "builds predating that check" aside too, since today that is every build and it read as though the check were already deployed.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
pllesperance-lc: Fixed in ff7ae58f: it is now a `#### When a cloud service can't be read` section (matching the page's existing `####` style), the "defaults to the full suite" duplication with the `events.cloud_services` bullet is gone, and the narrow-the-list remedy is stated once, up front, instead of three times.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
pllesperance-lc: Fixed in ff7ae58f: the all-services-refused escalation is now a third bullet in that admonition, noting that each service still warns individually and the error is raised on top of them, once per episode. Retitled to "When a 403 is reported as an error instead" so the list reads as exhaustive over exactly that question.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
```diff
@@ -219,7 +239,21 @@ harmony:
...
+        - Correct: `https://cloudinfra-gw-eu.portal.checkpoint.com`
+        - Incorrect: `https://cloudinfra-gw-eu.portal.checkpoint.com/auth/external`
+
+        Get this wrong and the adapter fails on its very first call, authentication, with the segment repeated:
+
+        ```
+        Cannot POST /auth/external/auth/external
+        ```
+
+        That is the only symptom you will see — authentication runs before anything else, so the data calls never get far enough to fail on their own mangled paths.
```
pllesperance-lc: With the startup check this page documents at L31, this symptom is unreachable. `https://cloudinfra-gw-eu.portal.checkpoint.com/auth/external` fails `HarmonyConfig.Validate` -> `validateGatewayURL` (`indexPathSegment` matches `/auth/external` as a whole segment), so `NewHarmonyAdapter` returns an error and the adapter never makes the auth call. `Cannot POST /auth/external/auth/external` is the pre-#320 behaviour — the adapter's own README says so explicitly: "that used to be accepted and then 404 every request as POST /auth/external/auth/external, and is now rejected by Validate with an error naming the value to use instead". The closing sentence ("That is the only symptom you will see — authentication runs before anything else...") compounds it, since there is no request at all.

Replace the paragraph, the code block and the closing sentence with the startup rejection the operator will actually see:

> Get this wrong and the adapter refuses to start, with an error naming the value to use instead:
>
> ~~~
> url: must be the gateway base URL, but "https://cloudinfra-gw-eu.portal.checkpoint.com/auth/external" already contains the API path "/auth/external", which the adapter appends itself — use "https://cloudinfra-gw-eu.portal.checkpoint.com" instead
> ~~~

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
```diff
@@ -41,7 +41,27 @@ All duration fields below are parsed with [`time.ParseDuration`](https://pkg.go.
 - `events.page_limit` *(optional)*: page size for the records-retrieval API. Defaults to `100`. The gateway rejects values below `10` with HTTP 400.
 - `events.limit` *(optional)*: cap on records returned per cloud service per poll. Defaults to `5000`.
 
-If a configured `cloud_service` is not provisioned for the tenant the gateway returns the query in state `Canceled`; the adapter logs one warning per poll and keeps going (it does not surface as an error). Remove the service from `cloud_services` to silence the warning.
+#### When a cloud service can't be read
+
+`events.cloud_services` defaults to the full Harmony suite, so a tenant licensed for only part of it will be queried for products it cannot read. Set `events.cloud_services` to just the products the tenant is licensed for and the rest of this section stops applying.
```
pllesperance-lc: "the rest of this section stops applying" is false for two of the cases this same section goes on to describe:

- L55, an API key missing the *Logs as a Service* grant: every cloud service is refused regardless of how the list is narrowed, and that is the case that leaves the source shipping nothing.
- L64, a misspelled service name: a hard error every poll, and narrowing the list is precisely when the names get hand-typed, so this advice makes it *more* likely, not less.

Narrow the claim to the case it actually covers, e.g.:

> Set `events.cloud_services` to just the products the tenant is licensed for — being careful to spell them exactly as below — and the licensing case stops applying.

pllesperance-lc:
Fifth round. Two open findings:

- L250-256: `Cannot POST /auth/external/auth/external` is unreachable — `validateGatewayURL` rejects that `url` before the adapter authenticates, so the symptom is the startup error, not a 404.
- L46: "the rest of this section stops applying" is false for the missing *Logs as a Service* grant (L55) and the misspelled name (L64), neither of which narrowing the list fixes.

Non-blocking: `events.cloud_services` deduplication and the escaped-path matching detail are invisible to operators and don't need a mention; L58's "only ... is soft-failed" reads as excluding `Canceled` until the admonition title is taken as scope; the all-refused latch counts only 403s, not `Canceled`, which L48's "refuses ... in one of two ways" slightly blurs; the pre-existing L41 note that the gateway rejects `page_limit` below 10 is unobservable since `Validate` clamps to 10. Rendering verified (admonition nests correctly in list item 5; `toc_depth: 3` keeps the new `####` out of the nav), and nothing internal is leaked.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
pllesperance-lc: Fixed in e23667e2. Correct, and it was my own overcorrection — the previous round fixed the wrong claim that every request doubles the suffix, and replaced it with a symptom the startup check makes unreachable. The admonition now shows the actual Validate error, which names the base URL to use, and keeps the 404 only as what an adapter released before the check does. Worth retaining in that form: it is the string someone hitting this on an older build will search for.

docs/2-sensors-deployment/adapters/types/checkpoint-harmony.md
pllesperance-lc: Fixed in e23667e2: the sentence now says narrowing the list fixes that case and its warnings, but explicitly not the two below it that survive it — a key without the Logs as a Service service, and a misspelled name, which narrowing makes more likely rather than less.

pllesperance-lc:
Parking this alongside refractionPOINT/usp-adapters#320, which it documents — that PR is closed for now too, so this page would describe behaviour that isn't shipping.

Closing to be revisited. The branch `solve/harmony-auth-url-docs-h403` is left in place, so reopening restores the work and its review history intact. CI was green, including the rendered-docs build.

Worth noting if this is picked up separately: the `/auth/external` copy-paste trap in the setup steps is a live documentation bug independent of the adapter change, and is what caused the original report.

PR #426, opened 2026-09-17
Email Security: forcing a remediation in alert-only mode

maximelb:
## Summary

An organization with no automation in `enforce` mode is in alert-only mode, and that withholds **every** remediation — including one a person starts from the console, CLI, API or an AI agent. Several pages said the opposite ("`alert_only` withholds automation, not people"); this corrects them and documents the explicit per-action `force` override.

- **remediation.md** — new *Forcing an action in alert-only mode* section: how each caller forces (console, CLI `--force`, SDK `force=True`, REST `force`, AI agents, D&R `force: true`), boolean-only semantics, not part of confirmation tokens, forced bulk execute is a new job, audit row + `EMAIL_ACTION` `forced: true`, and a D&R rule to alert on overrides. Bulk execute/status fields `force` / `force_required`.
- **api-reference.md** — `force` request field on the three action routes, `force_required` responses, bulk status fields.
- **cli.md**, **messages.md**, **campaigns.md**, **automation.md**, **ai-triage.md**, **getting-started.md**, **policy.md**, **index.md** — stale "people always execute" claims corrected; `--force` / `force: true` examples.
- **policy.md / troubleshooting.md** — a disabled `mailsec_policy` record is not applied and is reported as an org error (component `mailsec/policy`, about once a day).
- **custom-rules.md** — `name` and `fp_notes` are required on `dr-mail` rules (a save without them is refused); the `post_verdict` example now carries both, plus a `force: true` example.

`--force` needs a CLI release that includes it; the PR should go out with the product release.

## Checks

- `mkdocs build --strict` passes
- `markdownlint-cli2` on docs: 0 issues


maximelb:
merge with the prod release

maximelb:
## Adversarial self-review (pristine `git archive` of 4c38f6f, plus fix-up b96115e)

This repo does not use `/lc-review`, so the diff was reviewed from a clean extraction against origin/master.

**Accuracy vs the shipped behaviour**
- Alert-only withholds every remediation, whoever starts it: all "withholds automation, not people" / "always executes" claims are gone (re-grepped `docs/email-security`: none left).
- `force` on the three action routes, boolean `true` only, not part of confirmation tokens, previews don't take it: matches the gateway and backend.
- `force_required`: single action when `result == alert_only`; campaign sweep when `alert_only > 0`; bulk status `force` + `force_required`; forced bulk execute gets a new `bulk_id`: matches.
- Audit: a forced action gets its own row, `request.force: true`, and `EMAIL_ACTION` `forced: true` only when forced. The override-alert rule uses `event/forced`, following the `event/action` / `event/result` convention of the existing EMAIL_ACTION examples.
- CLI `--force` on `message action`, and on the `bulk-action` / `campaign action` execute (a preview prints a note), plus SDK `force=True` on `act_on_message` / `act_on_campaign` / `bulk_action_execute`: matches the merged CLI change. It isn't in a published CLI release yet, so the docs say it needs a recent release. Hence "merge with the prod release".
- A forced `banner_message` is performed even when the banners record is disabled: matches the executor (force is checked before the banner consent).
- A disabled `mailsec_policy` record raises an org error with component `mailsec/policy` and the quoted text, paced to about once a day: matches.
- `dr-mail`: `name` and `fp_notes` are both required on save (the hive validator refuses either missing). Fixed the Fields table and added both to the `post_verdict` example. The other `dr-mail` examples already had them.
- No page mentioned an optional-quota deferral for remediation, so nothing changed there.

**Found and fixed during review**
- api-reference: the `force` field on `POST /messages/{msg_uuid}/actions` linked to `#action-results` instead of the new `#forcing-an-action` subsection. Fixed in b96115e.

**Public-repo hygiene**
- No private repo names, internal PR references, internal URLs, org IDs, credentials or PII. No other email-security vendor named.

**Checks**
- `mkdocs build --strict`: passes. `markdownlint-cli2`: 0 issues.

No open issues.

lcbill (approved):

maximelb:
**Held — the backend is live, but the console and the CLI this page tells customers to use are not released.**

The backend half verifies cleanly against `legion_mailsec`, `go-mailsec` and `ext-email-security`, and that code is in production (`v0.1.0-mailsec131`, all six regions): the single choke point in `remediate.go`, `force_required` on a withheld action, boolean-`true`-only forcing at all four layers, `force` excluded from both confirmation tokens, a forced bulk execute running as a new `bulk_id`, the forced action as its own audit row with `force: true` in its `request`, `forced: true` on `EMAIL_ACTION` and absent otherwise, the extension's `force` field, and the `mailsec/policy` disabled-record error — that string matches character for character.

Two client surfaces do not back the page yet:

- **The console.** The force UI landed on `web-app-frontend` master today in #4413. It is **not** in tag `6.4.0` (cut 2026-09-16) and it is **not in the deployed bundle** — `forceRequired` is absent from the live `app.limacharlie.io` assets, while `judgedVia` and `laneUnavailableHelp` (both in 6.4.0) are present. So the *"Quarantine anyway?"* row describes something a customer opening the console today cannot do. What they actually see is the withheld sentence with no override.
- **The CLI and SDK.** `--force` and `force=True` exist only on `python-limacharlie` master. PyPI `limacharlie` 5.6.2 ships **no `mailsec` module at all** — no commands, no SDK methods — and the CHANGELOG has no `--force` entry. The PR's closing caveat ("needs a recent `limacharlie` release") understates this: there is no released CLI with the command in any form.

Two smaller precision points for whenever it does merge:

- `remediate.go:118` exempts `source: analyst` from the alert-only check unconditionally. It is unreachable from any live caller today (the gateway stamps `api` for every public API/console/CLI call, and body-supplied `source` is not forwarded), so the page is true in practice — but "withholds **every** action" is not literally what the code says.
- `force_required` is always present on an action response, `false` on `ok`/`skipped`/`failed` — not omitted except when withheld.
- The disabled-policy notice is paced **per pod**, so an org with N collectors can see up to N notices a day, and a restart resets the clock. "About once a day" is right per pod.

Suggest re-checking once the console ships and the CLI is on PyPI; the text itself is in good shape.

